PR #16, opened 2017-12-01
bump version to 0.1.0

testforstephen:
Signed-off-by: Jinbo Wang <jinbwan@microsoft.com>

### DIFF
--- a/extension/package.json
+++ b/extension/package.json
@@ -5,7 +5,7 @@
     "repository": {
         "url": "https://github.com/Microsoft/vscode-java-test"
     },
-    "version": "0.0.1",
+    "version": "0.1.0",
     "publisher": "vscjava",
     "bugs": {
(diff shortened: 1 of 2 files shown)


extension/package.json
```diff
@@ -5,7 +5,7 @@
     "repository": {
         "url": "https://github.com/Microsoft/vscode-java-test"
     },
-    "version": "0.0.1",
+    "version": "0.1.0",
```
yaohaizh: Should the internal java package also need changes?

ansyral:
leave out the change of package.lock.json?

extension/package.json
testforstephen: java package has already been changed to 0.1.0.

ansyral: already updated

PR #19, opened 2017-12-01
add package find-java-home as last resort

ansyral:

i suppose we could save time to call `find-java-home` directly, but i'm not sure whether there is any potential bug that vscode-java extension would do some check themselves and use find-java-home as last resort, let's follow vscode-java 's behavior first. I would update the logic after the release since by then we have plenty time to test

### DIFF
--- a/extension/src/extension.ts
+++ b/extension/src/extension.ts
@@ -6,6 +6,7 @@
 import * as archiver from 'archiver';
 import * as cp from 'child_process';
 import * as expandHomeDir from 'expand-home-dir';
+import * as findJavaHome from 'find-java-home';
 import * as fileUrl from 'file-url';
 import * as fs from 'fs';
 import * as glob from 'glob';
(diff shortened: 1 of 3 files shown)


extension/src/extension.ts
```diff
@@ -94,17 +95,18 @@ function checkJavaHome(): Promise<string> {
...
-        javaHome = expandHomeDir(javaHome);
-        if (!pathExists.sync(javaHome)) {
-            reject();
-        }
-        if (!pathExists.sync(path.resolve(javaHome, 'bin', JAVAC_FILENAME))) {
-            reject();
+        if (javaHome) {
+            javaHome = expandHomeDir(javaHome);
+            if (pathExists.sync(javaHome) && pathExists.sync(path.resolve(javaHome, 'bin', JAVAC_FILENAME))) {
+                resolve(javaHome);
```
testforstephen: `return resolve(javaHome);`

PR #28, opened 2017-12-01
change fixed port to random port

ansyral:

when port is fixed, launching multiple cases sequentially.  might hit the bp of wrong test case


extension/src/extension.ts
```diff
@@ -122,7 +123,7 @@ async function runTest(javaHome: string, tests: TestSuite[] | TestSuite, storage
     const suites = testList.map((s) => s.test);
     const uri = vscode.Uri.parse(testList[0].uri);
     const classpaths = classPathManager.getClassPath(uri);
-    const port = readPortConfig();
+    const port = await Utility.getFreePort();
```
testforstephen: try ... catch for getFreePort to handle the failure case.

akaroml: +1

Issue #23, opened 2017-12-01
Should not prompt save when closing test log


akaroml:
Test logs should be able to be closed without asking users to save the logs.

This can be achieved by using the DocumentContentProvider  API.

Issue #7, opened 2017-11-30
Should always save class test result file in multiple files with the same name?


bsaby:
Sample:
Method test result file:
![image](https://user-images.githubusercontent.com/22470312/33428004-3c4209ca-d602-11e7-8977-49529574f28f.png)
Class test result file:
![image](https://user-images.githubusercontent.com/22470312/33428016-48bd1578-d602-11e7-8dd0-1b7d5cd86d7f.png)

both of them have same file name, when i save files of them at the same time, Is it reasonable to always save the class test result file?

Issue #1, opened 2017-11-28
Can't save test result file in vscode


bsaby:
##### Environment
- Operating System: win10 x64
- JDK version: 1.18.0_152 x64
- Visual Studio Code version: 1.18.1 x64
- Java test extension version: #22

![image](https://user-images.githubusercontent.com/22470312/33311542-0b692070-d460-11e7-9a13-0a167ec76993.png)

bsaby:
The file can be saved now.
But an error occurs when i click status icon after saving test result file. 
![image](https://user-images.githubusercontent.com/22470312/33424517-07e2a94c-d5f7-11e7-85e2-2dc0e81085f2.png)

bsaby:
Now, the test report is read-only, so close it.

PR #48, opened 2017-12-05
use contentprovider API to load readonly test report

ansyral:

no need to save/not save test report any more.

### DIFF
--- a/extension/src/extension.ts
+++ b/extension/src/extension.ts
@@ -22,12 +22,13 @@ import * as vscode from 'vscode';
 import { Commands, Configs, Constants } from './commands';
 import { JUnitCodeLensProvider } from './junitCodeLensProvider';
 import { TestResourceManager } from './testResourceManager';
-import { OutputChannel, SnippetString, ExtensionContext, window } from 'vscode';
+import { OutputChannel, ExtensionContext, Uri } from 'vscode';
 import { TestSuite, TestLevel } from './protocols';
 import { ClassPathManager } from './classPathManager';
 import { TestResultAnalyzer } from './testResultAnalyzer';
 import TelemetryReporter from 'vscode-extension-telemetry';
 import { Logger, LogLevel } from './logger';
+import TestReportProvider, { encodeTestSuite } from './testReportProvider';
 
 const isWindows = process.platform.indexOf('win') === 0;
 const JAVAC_FILENAME = 'javac' + (isWindows ? '.exe' : '');
@@ -40,10 +41,12 @@ let running: boolean = false;
 
 // this method is called when your extension is activated
 // your extension is activated the very first time the command is executed
-export function activate(context: vscode.ExtensionContext) {
+export function activate(context: ExtensionContext) {
     activateTelemetry(context);
     const codeLensProvider = new JUnitCodeLensProvider(onDidChange, testResourceManager, logger);
     context.subscriptions.push(vscode.languages.registerCodeLensProvider(Constants.LANGUAGE, codeLensProvider));
+    const testReportProvider: TestReportProvider = new TestReportProvider(testResourceManager);
+    context.subscriptions.push(vscode.workspace.registerTextDocumentContentProvider(TestReportProvider.scheme, testReportProvider));
 
     vscode.workspace.onDidChangeTextDocument((event) => {
         const uri = event.document.uri;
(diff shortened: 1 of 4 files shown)


extension/src/extension.ts
```diff
@@ -22,12 +22,13 @@ import * as vscode from 'vscode';
...
 import { JUnitCodeLensProvider } from './junitCodeLensProvider';
 import { TestResourceManager } from './testResourceManager';
-import { OutputChannel, SnippetString, ExtensionContext, window } from 'vscode';
+import { OutputChannel, ExtensionContext, Uri } from 'vscode';
 import { TestSuite, TestLevel } from './protocols';
 import { ClassPathManager } from './classPathManager';
 import { TestResultAnalyzer } from './testResultAnalyzer';
 import TelemetryReporter from 'vscode-extension-telemetry';
 import { Logger, LogLevel } from './logger';
+import TestReportProvider, { encodeTestSuite } from './testReportProvider';
```
yaohaizh: Should avoid use the mixed import style.

ansyral:
#1 #7 #23

extension/src/extension.ts
```diff
@@ -22,12 +22,13 @@ import * as vscode from 'vscode';
 import { Commands, Configs, Constants } from './commands';
 import { JUnitCodeLensProvider } from './junitCodeLensProvider';
 import { TestResourceManager } from './testResourceManager';
-import { OutputChannel, SnippetString, ExtensionContext, window } from 'vscode';
+import { OutputChannel, ExtensionContext, Uri } from 'vscode';
```
yaohaizh: Already import the vscode as *.  This import is not necessary.

extension/src/extension.ts
```diff
@@ -218,31 +220,12 @@ async function runSingleton(javaHome: string, tests: TestSuite[] | TestSuite, st
...
 function showDetails(test: TestSuite) {
     const editor = vscode.window.activeTextEditor;
-    const dir = path.dirname(editor.document.uri.fsPath);
-    const uri = vscode.Uri.parse(`${Constants.TEST_OUTPUT_SCHEME}:${path.join(dir, 'test-result-' + test.test)}`);
+    const uri: Uri = encodeTestSuite(editor.document.uri, test);
     return vscode.workspace.openTextDocument(uri).then(doc => {
-        return vscode.window.showTextDocument(doc, editor.viewColumn + 1).then(edit => {
-            edit.insertSnippet(new SnippetString(getTestReport(test)), new vscode.Position(0, 0));
-        });
+        return vscode.window.showTextDocument(doc, editor.viewColumn + 1);
```
yaohaizh: Suggestion, simple sentence  can remove return and {}

extension/src/testReportProvider.ts
```diff
@@ -0,0 +1,57 @@
+import { TestResourceManager } from "./testResourceManager";
+import { TestSuite, TestLevel } from "./protocols";
+import { TextDocumentContentProvider, Uri } from "vscode";
+
+import * as os from 'os';
```
yaohaizh: Should add eslint files for checking styles

extension/src/testReportProvider.ts
```diff
@@ -0,0 +1,57 @@
...
+    }
+}
+
+export function encodeTestSuite(uri: Uri, test: TestSuite): Uri {
+    const query = JSON.stringify([uri.toString(), test.test.replace('#', '%23')]);
+    return Uri.parse(`${TestReportProvider.scheme}:test-report?${query}`);
+}
+
+export function decodeTestSuite(uri: Uri): [Uri, string] {
+    let [target, test] = <[string, string]>JSON.parse(uri.query);
```
yaohaizh: const

extension/src/extension.ts
ansyral: I added this to not introduce `vscode.` frequently. Removing `windows` as it is lowercase and might be confused with local variables. I think it is good to keep consistent. Which way do you think is better? `import * as vscode from vscode` or import all the used type in vscode module explicitly

yaohaizh: Either import as * or import separate. Either way works but not mixed.

extension/src/testReportProvider.ts
ansyral: already added tslint.json at the end of last release, however I didn't make it too strict as it introduced many errors. might need to open a separate pr to update tslint.json and fix all style errors.

extension/src/testReportProvider.ts
```diff
@@ -0,0 +1,57 @@
+import { TestLevel, TestSuite } from "./protocols";
+import { TestResourceManager } from "./testResourceManager";
+
+import * as os from 'os';
+import { TextDocumentContentProvider, Uri } from "vscode";
+
+export class TestReportProvider implements TextDocumentContentProvider {
+
+    static scheme = 'test-report';
```
yaohaizh: Const?

ansyral: const isn't allowed as class member

PR #51, opened 2017-12-06
format test report

ansyral:
Signed-off-by: xuzho <xuzho@microsoft.com>

PR #57, opened 2017-12-15
use template files to render

ansyral:

### DIFF
--- a/extension/resources/templates/report_class.liquid
+++ b/extension/resources/templates/report_class.liquid
@@ -1,0 +1,86 @@
+<head>
+    <meta http-equiv="Content-Security-Policy" content="default-src 'none'; img-src *; style-src 'nonce-self'; script-src 'nonce-check';">
+    <style nonce="self">
+        .run {
+            width:33%;
+            float:left;
+        }
+        .failed {
+            width:33%;
+            float:left;
+        }
+        .passed {
+            width:33%;
+            float:left;
+        }
+        #status div span {
+            display: table-cell;
+        }
+        #status div .counter {
+            background-color:rgb(200,200,200);
+            color:black;
+            width:100%;
+        }
+        #color-status {
+            background-color:
+            {% if failedCount > 0 %}
+                red;
+            {% elsif (passCount | plus: skippedCount) == totalCount %}
+                green;
+            {% else %}
+                orange;
+            {% endif %}
+            height: 15px;
+            float:left;
+            width:100%;
+        }
+    </style>
+</head>
+{% assign children = allTests %}
+{% assign passedChildren = passedTests %}
+{% assign failedChildren = failedTests %}
+<body>
+    <div>Test report of class: {{test}}</div>
+    <hr/>
+    <div id="status">
+        <div class="run">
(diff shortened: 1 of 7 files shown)


ansyral:
#51

extension/resources/templates/report_class.liquid
```diff
@@ -0,0 +1,86 @@
...
+    {% endif %}
+    <div id="filter">
+        <label><input type="radio" name="test" value="all" checked ="checked"> All tests</label>
+        <label><input type="radio" name="test" value="passed"> Passed</label>
+        <label><input type="radio" name="test" value="failed"> Failed</label>
+    </div>
+    <div id="children">
+        {% include 'report_method_table' %}
+    </div>
+    <script nonce="check">
```
akaroml: Remember to get rid of the front-end logic.

extension/resources/templates/report_class.liquid
```diff
@@ -0,0 +1,86 @@
+<head>
+    <meta http-equiv="Content-Security-Policy" content="default-src 'none'; img-src *; style-src 'nonce-self'; script-src 'nonce-check';">
+    <style nonce="self">
```
akaroml: You can create a bootstrap stylesheet and reference it in a minute. Goto getbootstrap.com,

extension/resources/templates/report_failed_method_table.liquid
```diff
@@ -0,0 +1,42 @@
+<table border="1">
```
akaroml: Table is not the recommended way of displaying data. Use div instead when you have bootstrap ready.

extension/resources/templates/report_class.liquid
```diff
@@ -0,0 +1,86 @@
+<head>
```
akaroml: Just some opinion that the liquid template is not syntactically condensed. Here's an example of Jade:
```jade
doctype html
html(lang="en")
  head
    title= pageTitle
    script(type='text/javascript').
      if (foo) bar(1 + 5)
  body
    h1 Jade - node template engine
    #container.col
      if youAreUsingJade
        p You are amazing
      else
        p Get on it!
      p.
        Jade is a terse and simple templating language with a
        strong focus on performance and powerful features.
```

Not sure why you chose to use liquid-node which only has 698 downloads last week while jade has 535,984...

extension/resources/templates/report_class.liquid
```diff
@@ -0,0 +1,86 @@
...
+    <div id="color-status">
+        &nbsp;
+    </div>
+    {% if result and result.summary %}
+        <div id="summary">
+            <span>Summary: {{ result.summary }}</span>
+        </div>
+    {% endif %}
+    <div id="filter">
+        <label><input type="radio" name="test" value="all" checked ="checked"> All tests</label>
```
akaroml: Try use anchor.

extension/resources/templates/report_class.liquid
ansyral: I haven't heard about Jade before. After some investigation, Jade's API doesn't support async and it cannot load templates from a directory, however it supports more complex syntax than liquid. since jade is more popular and may be better maintained, switch to jade is ok.

Issue #46, opened 2017-12-05
Show test statistics in status bar


akaroml:
Currently, we cannot see the numbers of passed or failed cases. This can be displayed inside the status bar.

bsaby:
It would be better to show progress when loading test explorer in status bar

Issue #41, opened 2017-12-05
Provide switch to optionally show output window


akaroml:
Currently, we force showing output window whenever users run test cases. This could be disruptive when users are currently working on debug console or integrated terminal.

We can provide an option in user settings to allow switch that behavior on and off.

PR #59, opened 2017-12-20
add test status bar

ansyral:

#46 #41

### DIFF
--- a/extension/src/extension.ts
+++ b/extension/src/extension.ts
@@ -17,7 +17,8 @@ import * as os from 'os';
 import * as path from 'path';
 import * as pathExists from 'path-exists';
 import * as rimraf from 'rimraf';
-import { commands, debug, languages, window, workspace, EventEmitter, ExtensionContext, OutputChannel, Uri, ViewColumn } from 'vscode';
+// tslint:disable-next-line
+import { commands, debug, languages, window, workspace, EventEmitter, ExtensionContext, OutputChannel, ProgressLocation, Uri, ViewColumn } from 'vscode';
 import TelemetryReporter from 'vscode-extension-telemetry';
 
 import { ClassPathManager } from './classPathManager';
@@ -176,7 +177,7 @@ async function runTest(javaHome: string, tests: TestSuite[] | TestSuite, storage
     }
 
     const testResultAnalyzer = new TestResultAnalyzer(testList);
-    await new Promise((resolve, reject) => {
+    await testStatusBarItem.update(testList, new Promise((resolve, reject) => {
         let error: string = '';
         const process = cp.exec(params.join(' '));
         process.on('error', (err) => {
@@ -195,7 +196,6 @@ async function runTest(javaHome: string, tests: TestSuite[] | TestSuite, storage
         process.on('close', () => {
             testResultAnalyzer.feedBack();
             onDidChange.fire();
-            testStatusBarItem.update(testList);
             rimraf(storageForThisRun, (err) => {
                 if (err) {
                     logger.logError(`Failed to delete storage for this run. Storage path: ${err}`);
@@ -221,7 +221,7 @@ async function runTest(javaHome: string, tests: TestSuite[] | TestSuite, storage
                 });
             }, 500);
         }
-    });
+    }));
 }
 
 async function runSingleton(javaHome: string, tests: TestSuite[] | TestSuite, storagePath: string, isDebugMode: boolean) {
--- a/extension/src/extension.ts
+++ b/extension/src/extension.ts
@@ -48,7 +48,7 @@ let running: boolean = false;
 // your extension is activated the very first time the command is executed
 export async function activate(context: ExtensionContext) {
     activateTelemetry(context);
-    testStatusBarItem.init(testResourceManager.refresh());
+    await testStatusBarItem.init(testResourceManager.refresh());
     const codeLensProvider = new JUnitCodeLensProvider(onDidChange, testResourceManager, logger);
     context.subscriptions.push(languages.registerCodeLensProvider(Configs.LANGUAGE, codeLensProvider));
     const testReportProvider: TestReportProvider = new TestReportProvider(context, testResourceManager);
--- a/extension/src/extension.ts
+++ b/extension/src/extension.ts
@@ -30,12 +30,14 @@ import { TestLevel, TestSuite } from './protocols';
 import { encodeTestSuite, parseTestReportName, TestReportProvider } from './testReportProvider';
 import { TestResourceManager } from './testResourceManager';
 import { TestResultAnalyzer } from './testResultAnalyzer';
+import { TestStatusBarSingleton } from './testStatusBarSingleton';
 import { TestExplorer } from './Explorer/testExplorer';
 import { TestTreeNode } from './Explorer/testTreeNode';
 
 const isWindows = process.platform.indexOf('win') === 0;
 const JAVAC_FILENAME = 'javac' + (isWindows ? '.exe' : '');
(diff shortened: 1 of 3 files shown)


extension/package.json
```diff
@@ -29,12 +29,20 @@
...
     ],
     "main": "./out/src/extension",
     "contributes": {
         "javaExtensions": [
             "./server/com.microsoft.java.test.plugin-0.1.0.jar"
+        ],
+        "commands": [
+            {
+                "command": "java.test.show.output",
+                "title": "Show Test Output pannel",
```
akaroml: remove "panel"

extension/src/extension.ts
```diff
@@ -62,6 +64,8 @@ export function activate(context: ExtensionContext) {
             withScopeAsync(() => runSingleton(javaHome, suites, context.storagePath, true), "Debug Test")));
         context.subscriptions.push(commands.registerCommand(Commands.JAVA_TEST_SHOW_DETAILS, (test: TestSuite) =>
             withScopeAsync(() => showDetails(test), "Show Test details")));
+        context.subscriptions.push(commands.registerCommand(Commands.JAVA_TEST_SHOW_OUTPUT, () =>
+            withScopeAsync(() => outputChannel.show(), "Show Test output panel")));
```
akaroml: Try use constants for the string. And remove "panel".

akaroml: "Show test output"

extension/package.json
akaroml: "Show test output"

extension/src/testStatusBar.ts
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+import { window, StatusBarAlignment, StatusBarItem } from "vscode";
+
+import * as Commands from "./commands";
+import { TestLevel, TestStatus, TestSuite } from "./protocols";
+
+export class TestStatusBar {
```
akaroml: TestStatusBarItem

extension/src/testStatusBar.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export class TestStatusBar {
+    private statusBar: StatusBarItem = window.createStatusBarItem(StatusBarAlignment.Left, Number.MIN_VALUE);
+
+    public update(tests: TestSuite[]): void {
+        let failedCount: number = 0;
+        let passedCount: number = 0;
+        const testMethods: TestSuite[] = tests.map((t) => t.level === TestLevel.Method ? [t] : t.children)
+                                 .reduce((a, b) => a.concat(b));
+        for (const t of testMethods) {
```
akaroml: failedCount = testMethods.reduce(m => t.result.status === TestStatus.Fail ? 1 : 0);

extension/src/testStatusBar.ts
```diff
@@ -0,0 +1,35 @@
...
+                continue;
+            }
+            if (t.result.status === TestStatus.Fail) {
+                failedCount++;
+            } else if (t.result.status === TestStatus.Pass) {
+                passedCount++;
+            }
+        }
+        this.statusBar.text = '$(x)' + failedCount + '  $(check)' + passedCount;
+        this.statusBar.command = Commands.JAVA_TEST_SHOW_OUTPUT;
```
akaroml: Why this leads to the output instead of the corresponding test report?

extension/src/testStatusBar.ts
```diff
@@ -0,0 +1,35 @@
...
+            if (t.level !== TestLevel.Method || !t.result) {
+                continue;
+            }
+            if (t.result.status === TestStatus.Fail) {
+                failedCount++;
+            } else if (t.result.status === TestStatus.Pass) {
+                passedCount++;
+            }
+        }
+        this.statusBar.text = '$(x)' + failedCount + '  $(check)' + passedCount;
```
yaohaizh: Using string literals

extension/src/testStatusBar.ts
yaohaizh: This should be a singleton. Make sure it cannot be created more than once.

extension/src/testStatusBar.ts
ansyral: later on, we would support triggering tests from test explorer, we don't have a summary report for a trigger, while output and each trigger are 1:1 match, so i think directing to output panel is better

extension/src/testStatusBar.ts
ansyral: nice

extension/src/testStatusBar.ts
ansyral: `map` and `reduce` makes code neat, but for the scenario, we have to loop twice, however using for-loop we only need one loop.

extension/src/extension.ts
```diff
@@ -30,10 +30,12 @@ import { TestLevel, TestSuite } from './protocols';
 import { encodeTestSuite, TestReportProvider } from './testReportProvider';
 import { TestResourceManager } from './testResourceManager';
 import { TestResultAnalyzer } from './testResultAnalyzer';
+import { TestStatusBarSingleton } from './testStatusBarSingleton';
```
yaohaizh: it is uncommon to use singleton as file name suffix.

ansyral:
i rethink about the workflow to show report when clicking on statusbar, maybe it is better to make all the reports the same format, namely the interface accepts a list of tests, no matter it is a test class, a test method or a list of test methods. Let me update this in another pr

extension/src/testStatusBarSingleton.ts
```diff
@@ -0,0 +1,67 @@
...
+                continue;
+            }
+            if (t.result.status === TestStatus.Fail) {
+                failedCount++;
+            } else if (t.result.status === TestStatus.Pass) {
+                passedCount++;
+            }
+        }
+        this.statusBarItem.text = `$(x)${failedCount}  $(check)${passedCount}`;
+        this.statusBarItem.command = Commands.JAVA_TEST_SHOW_OUTPUT;
```
akaroml: Still showing output window. Please show report instead.

extension/src/testStatusBarSingleton.ts
```diff
@@ -0,0 +1,67 @@
...
+            (reason) => {
+                this.statusBarItem.text = 'Failed to load tests';
+            });
+        });
+    }
+
+    public update(tests: TestSuite[], action: Thenable<void>) {
+        return window.withProgress({ location: ProgressLocation.Window }, async (p) => {
+            p.report({message: 'Running/Debugging test...'});
+            this.statusBarItem.text = 'View test output';
```
akaroml: Now the `View test output` is a new button beside the progress info. Can we just make the progress info a button? By doing so, users can just click the progress to see the logs.

akaroml: And suggest `View test logs...`

extension/src/testStatusBarSingleton.ts
```diff
@@ -0,0 +1,67 @@
...
+            return action.then(null,
+            (reason) => {
+                this.statusBarItem.text = 'Failed to load tests';
+            });
+        });
+    }
+
+    public update(tests: TestSuite[], action: Thenable<void>) {
+        return window.withProgress({ location: ProgressLocation.Window }, async (p) => {
+            p.report({message: 'Running/Debugging test...'});
```
akaroml: `Running tests...` is enough.

extension/src/testStatusBarSingleton.ts
ansyral: [#Pending]this needs to refactor reportProvider, i'll split it to another pr

PR #64, opened 2017-12-21
Add dark and light report templates

akaroml:
@ansyral please use these templates according to users theme settings.


ansyral:
changed in #68

PR #68, opened 2017-12-27
test report: support dark/light theme

ansyral:

### DIFF
--- a/extension/.gitignore
+++ b/extension/.gitignore
@@ -3,3 +3,4 @@ node_modules
 *.vsix
 .vscode-test
 .vscode
+resources/templates/css/*.css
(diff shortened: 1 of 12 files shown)


ansyral:
In the new report, if user clicks on "Passed Tests", report would be rendered in another view column. i've opened an [issue](https://github.com/Microsoft/vscode/issues/40818) in vscode to see whether they support specifying viewcolumn in link href, if not , then we still need to change back to original design

akaroml:
1. The navbar should have a backgroud color instead of transparent
2. The blank line between test cases should be completely removed
3. The cursor on the test case line should be pointer(hand)
4. The tab name should be `ClassName - All`, `ClassName - Error`, `ClassName - Passed`, `MethodName`

ansyral:
split to #70 #71

PR #77, opened 2018-01-03
refactor report

ansyral:
1. unify report of all levels
2. when test finished, click on the status bar, would jump to the report of this run


extension/resources/templates/report.pug
```diff
@@ -27,10 +27,11 @@ html(lang="en")
                         .progress-bar.bg-success(role='progressbar', style='width:25%', aria-valuenow='15', aria-valuemin='0', aria-valuemax='100') #{passCount} Passed
```
akaroml: Can we hide the `All, Failed, passed` options when we are showing a single case result?

extension/src/Utils/commandUtility.ts
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+import * as util from 'util';
+import { commands } from 'vscode';
+
+export class CommandUtility {
+    public static get(command: string, args: any[]) {
```
akaroml: The name `get` is too generic. Please find a better to describe its purpose.

extension/resources/templates/report.pug
ansyral: sure, but for class level, even though it might only contains one method, still think it is necessary to show this as it is kind of a container

extension/src/Utils/commandUtility.ts
```diff
@@ -5,7 +5,7 @@ import * as util from 'util';
 import { commands } from 'vscode';
 
 export class CommandUtility {
-    public static get(command: string, args: any[]) {
+    public static getCommandWithArgs(command: string, args: any[]) {
```
akaroml: We need to delete the old command whenever a new command is generated.

PR #87, opened 2018-01-10
refactor logging I: add interface and move some files

ansyral:


extension/src/Utils/Logging/logListener.ts
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+export interface ILogListener {
```
akaroml: @Eskibear is working on a common logger interface. Please hold this and use the new interface instead.

PR #98, opened 2018-01-16
refactor runner

ansyral:


extension/src/Runner/testRunnerContext.ts
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+import { TestSuite } from "../protocols";
+
+export interface ITestRunnerContext {
```
akaroml: This class is actually the parameters for ITestRunner.run. Suggest:
Eliminate it and expose the class members as parameters of ITestRunner.run
or
Rename the class to ITestRunnerParameters.
I would suggest the former because it's more explicit contract-wise.

extension/src/Runner/testRunner.ts
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+import { ITestRunnerContext } from "./testRunnerContext";
+
+export interface ITestRunner {
+    run(context: ITestRunnerContext): undefined | Promise<undefined>;
```
akaroml: Returning `undefined` will cost the caller an `if` check to handle the return value differently. Why not just use Promise and reject when there's something wrong?

extension/src/Runner/testRunner.ts
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+import { ITestRunnerContext } from "./testRunnerContext";
+
+export interface ITestRunner {
```
akaroml: I don't see it necessary to use `interface` instead of `class`. In JavaScript, there is no built-in `interface`. It is introduced by TypeScript to enforce type checking. The type checking is ideal for check object shape, e.g. checking the object inside a remote call response.
But in our case, we only need polymorphism. Classes work perfectly for that purpose. Using interfaces is an overkill in my opinion.

extension/src/Utils/classPathUtility.ts
```diff
@@ -0,0 +1,60 @@
...
+import * as archiver from 'archiver';
+import * as fileUrl from 'file-url';
+import * as fs from 'fs';
+import * as mkdirp from 'mkdirp';
+import * as os from 'os';
+import * as path from 'path';
+
+export class ClassPathUtility {
+    // TODO: refactor logger, no need to pass in logger instance.
+    public static processLongClassPath(
```
akaroml: The name `processLongClassPath` is not explicit regarding its purpose. Code should be self-describing. By reading the implementation, `generateClassPathFile` is the appropriate name.

extension/src/Utils/classPathUtility.ts
```diff
@@ -0,0 +1,60 @@
...
+import { Logger } from '../logger';
+
+import * as archiver from 'archiver';
+import * as fileUrl from 'file-url';
+import * as fs from 'fs';
+import * as mkdirp from 'mkdirp';
+import * as os from 'os';
+import * as path from 'path';
+
+export class ClassPathUtility {
```
akaroml: Do you think the members of this class popular in the future? If there would be only one caller, then suggest putting the functions into the caller class directly.
The processing logic is to resolve a very specific problem which the JarFileTestRunner has. So to me, it's not common enough to form a utility class.

extension/src/Runner/testRunner.ts
akaroml: The interface is just too generic. It is not clear enough to the caller about what to provide and what to expect. Please rework this. You could have stages like `beforeRun`, `run` and `postRun` and each have clear definitions of what to implement.

extension/src/Runner/testRunner.ts
ansyral: because i have no default behavior for the run method. So it is abstract class versus interface, from my point, class is also a syntax sugar in js, why not interface?

extension/src/Runner/testRunner.ts
ansyral: make sense

extension/src/Runner/testRunnerContext.ts
ansyral: we do have a lot of context info of this run, like `transactionid`, `storagepath`, `port`, etc. Putting them directly as the parameters of the method would make the interface vulnerable to change. However, I agree that we could put `tests` and `isDebugMode` as parameters directly

extension/src/Utils/classPathUtility.ts
ansyral: The methods here are to solve a general problem that class path might be too long for command line to handle, so I extract them to a utility method.

extension/src/Runner/runnerResultStream.ts
```diff
@@ -0,0 +1,16 @@
...
+import { Readable } from "stream";
+
+export class RunnerResultStream extends events.EventEmitter implements IRunnerResultStreamEvent {
+    constructor(public readonly stderr: Readable, public readonly stdout: Readable) {
+        super();
+    }
+}
+
+interface IRunnerResultStreamEvent {
+    on(event: 'finish', listener: () => void): this;
```
akaroml: 'this' as return type. What does it mean?

extension/src/Runner/testRunnerEnvironment.ts
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+import { TestSuite } from "../protocols";
+
+export interface ITestRunnerEnvironment {
```
akaroml: 'Environment' is misleading. Suggest 'Parameters'

extension/src/Runner/testRunnerEnvironment.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export interface ITestRunnerEnvironment {
+    tests: TestSuite[];
+    isDebugMode: boolean;
+    storagePath: string;
+    port: number | undefined;
+    transactionId: string | undefined; // TODO: remove later after refactoring logger
+}
+
+export class JarFileTestRunnerEnvironment implements ITestRunnerEnvironment {
```
akaroml: This class does not have any member functions. If it's just for type checking, you can use another extending interface which has the extra 'classpathStr' member.

extension/src/Runner/testRunner.ts
```diff
@@ -0,0 +1,12 @@
...
+// Licensed under the MIT license.
+
+import { TestSuite } from "../protocols";
+import { RunnerResultStream } from "./runnerResultStream";
+import { ITestRunnerEnvironment } from "./testRunnerEnvironment";
+
+export interface ITestRunner {
+    setup(tests: TestSuite[], isDebugMode: boolean): Promise<ITestRunnerEnvironment>;
+    run(env: ITestRunnerEnvironment): Promise<RunnerResultStream>;
+    updateTestStatus(env: ITestRunnerEnvironment, result: RunnerResultStream): Promise<void>;
```
akaroml: Who will be calling this and what does this function do?

extension/src/Runner/runnerResultStream.ts
```diff
@@ -0,0 +1,16 @@
...
+import * as events from "events";
+import { Readable } from "stream";
+
+export class RunnerResultStream extends events.EventEmitter implements IRunnerResultStreamEvent {
+    constructor(public readonly stderr: Readable, public readonly stdout: Readable) {
+        super();
+    }
+}
+
+interface IRunnerResultStreamEvent {
```
akaroml: There are already events defined in the stderr and stdout streams. Does this mean those are not enough?

extension/src/Runner/runnerResultStream.ts
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+import * as events from "events";
+import { Readable } from "stream";
+
+export class RunnerResultStream extends events.EventEmitter implements IRunnerResultStreamEvent {
```
akaroml: What does this stream really do?

extension/src/Runner/testRunnerEnvironment.ts
ansyral: nice

extension/src/Runner/testRunner.ts
ansyral: I'd like to create a single entry which has a pool of different runners and expose a static method `run` to outside. and this entry would call all the interface methods of `ITestRunner` to run the tests and update the status. It would be like this:
```
const runner: ITestRunner = getMatchedRunner(....);
const env = await runner.setup(tests, isDebug);
await runner.run(env).then((stream) => runner.updateTestStatus(env, stream));
```

extension/src/Runner/runnerResultStream.ts
ansyral: it follows `events.EventEmitter`. However, maybe I should rename the interface to IRunnerResultStreamEventEmitter

extension/src/Runner/runnerResultStream.ts
ansyral: the stream contains running result. Because I'd like to design the running behavior in an asynchronous way, therefore I need to expose a stream as run result so that it can update test status while running

extension/src/Runner/runnerResultStream.ts
ansyral: yes, i'd like to mark when the running completes and when some errors occurred to the runner itself as well

extension/src/Runner/testRunnerParameters.ts
```diff
@@ -0,0 +1,16 @@
...
+export interface ITestRunnerParameters {
+    tests: ITestInfo[];
+    isDebugMode: boolean;
+    storagePath: string;
+    port: number | undefined;
+    transactionId: string | undefined; // TODO: remove later after refactoring logger
+}
+
+export interface IJarFileTestRunnerParameters extends ITestRunnerParameters {
+    classpathStr: string;
```
akaroml: Do we also need to jar file name?

extension/src/Runner/testModel.ts
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+import { Range } from "vscode";
+import { TestLevel, TestResult } from "../protocols";
+
+export interface ITestInfo {
```
akaroml: The members are part of TestSuite. Can we make that an interface that extends this one?

PR #141, opened 2018-03-20
fix bug: data sent to stdout might be cut off

ansyral:


extension/src/Runner/JarFileRunner/jarFileTestRunner.ts
```diff
@@ -14,6 +14,7 @@ import { JarFileRunnerResultAnalyzer } from "./jarFileRunnerResultAnalyzer";
 
 import * as cp from 'child_process';
 import * as getPort from "get-port";
+import * as os from 'os';
```
yaohaizh: Should use either `'` or `"` for string in the file. Please configurate the rule about the string in the eslint or add eslint.

PR #148, opened 2018-03-28
configIV: add test config manager

ansyral:

### DIFF
--- a/extension/src/testConfigManager.ts
+++ b/extension/src/testConfigManager.ts
@@ -1,0 +1,91 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+import * as fs from 'fs';
+import * as path from 'path';
+import { window, workspace, ExtensionContext, TextEditor } from 'vscode';
+
+import { ProjectInfo, ProjectManager } from './projectManager';
+import * as Configs from './Constants/configs';
+import { TestConfig } from './Models/testConfig';
+import * as Logger from './Utils/Logger/logger';
+
+export class TestConfigManager {
+    private readonly _configPath: string;
+    constructor(context: ExtensionContext, private _projectManager: ProjectManager) {
+        this._configPath = context.asAbsolutePath(Configs.TEST_LAUNCH_CONFIG_NAME);
+    }
+
+    public get configPath(): string {
+        return this._configPath;
+    }
+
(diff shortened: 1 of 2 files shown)


extension/src/testConfigManager.ts
```diff
@@ -0,0 +1,91 @@
...
+import { window, workspace, ExtensionContext, TextEditor } from 'vscode';
+
+import { ProjectInfo, ProjectManager } from './projectManager';
+import * as Configs from './Constants/configs';
+import { TestConfig } from './Models/testConfig';
+import * as Logger from './Utils/Logger/logger';
+
+export class TestConfigManager {
+    private readonly _configPath: string;
+    constructor(context: ExtensionContext, private _projectManager: ProjectManager) {
```
yaohaizh: Why not directly pass in configPath?

ansyral: I'd like it to be managed by TestConfigManager, right now it uses extension path, but it could be other places

extension/src/testConfigManager.ts
```diff
@@ -0,0 +1,91 @@
...
+            fs.access(this._configPath, (err) => {
+                if (err) {
+                    const config: TestConfig = this.createDefaultTestConfig();
+                    const content: string = JSON.stringify(config, null, 4);
+                    fs.writeFile(this._configPath, content, 'utf-8', (writeErr) => {
+                        if (writeErr) {
+                            Logger.error('Failed to create default test config!', {
+                                error: writeErr,
+                            });
+                            reject(writeErr);
```
yaohaizh: Should return directly; otherwise will be call resolve after reject

extension/src/testConfigManager.ts
```diff
@@ -0,0 +1,91 @@
...
+        });
+    }
+
+    public editConfig(): Promise<TextEditor> {
+        const editor = window.activeTextEditor;
+        return this.loadConfig().then(() => {
+            return workspace.openTextDocument(this._configPath).then((doc) => {
+                return window.showTextDocument(doc, editor.viewColumn);
+            }, (err) => {
+                Logger.error(err);
```
yaohaizh: The error have be already logged

extension/src/testConfigManager.ts
ansyral: For promise, the resolve call after reject would be ignored. But I would add return to avoid confusion

PR #151, opened 2018-03-28
add logic for config

ansyral:


extension/src/Runner/testRunnerWrapper.ts
```diff
@@ -121,4 +124,31 @@ export class TestRunnerWrapper {
...
+            process.on('error', (err) => {
+                Logger.error(
+                    `Error occurred while executing prelaunch task. Name: ${err.name}. Message: ${err.message}. Stack: ${err.stack}.`,
+                    {
+                        stack: err.stack,
+                    });
+                reject(err);
+            });
+            process.stderr.on('data', (data) => {
+                Logger.error(`Error occurred: ${data.toString()}`);
```
yaohaizh: Should output to let user know the process information of process?

ansyral: it would be output to output channel

Issue #155, opened 2018-03-29
Cannot read property 'viewColumn' of undefined


sallyur:
1. Open one project
2. Make sure there is no .java file opened
3. Ctrl+shift+p open command panel
4. Trigger command java: Edit test configuration
Error in toggle developer tools:
![image](https://user-images.githubusercontent.com/28860891/38081114-c99e867e-3375-11e8-8de2-baeef74478be.png)

PR #157, opened 2018-03-30
fix bug #155

ansyral:


extension/src/testConfigManager.ts
```diff
@@ -52,7 +52,7 @@ export class TestConfigManager {
         const editor = window.activeTextEditor;
         return this.loadConfig().then(() => {
             return workspace.openTextDocument(this._configPath).then((doc) => {
-                return window.showTextDocument(doc, editor.viewColumn);
+                return window.showTextDocument(doc, editor ? editor.viewColumn : undefined);
```
yaohaizh: For comment, it should clearly define what this is about, but not number or links. Indirect reference will make things unclear.

PR #162, opened 2018-04-12
kill runner if timeout

ansyral:


ansyral:
add cancel test instead

PR #166, opened 2018-04-16
add cancel test command into menu when user clicks status bar

ansyral:


extension/src/Constants/commands.ts
```diff
@@ -35,6 +35,10 @@ export const JAVA_TEST_OPEN_LOG = 'java.test.open.log';
 
 export const JAVA_CONFIGURE_TEST_COMMAND = 'java.test.configure';
 
+export const JAVA_CANCEL_TEST = 'java.test.cancel';
+
+export const JAVA_STATUS_COMPOSITE_COMMAND = 'java.status';
```
yaohaizh: This command without `test` is easily conflict with the vscode-java command.

PR #194, opened 2018-05-09
Move test explorer to Test view container

sandy081:
Hi,

I am developer from VS Code and recently we added a new test view container. Refer to the release notes [here](https://github.com/Microsoft/vscode-docs/blob/vnext/release-notes/v1_23.md#contributions-to-the-activity-bar). 

I am going through the popular extensions in the market place those contribute test related custom views and moving those view to Test view container.

I would request to accept this PR and publish the extension with changes.

Thanks

### DIFF
--- a/extension/package.json
+++ b/extension/package.json
@@ -21,7 +21,7 @@
     ],
     "aiKey": "90c182a8-8dab-45d4-bfb8-1353eb55aa7f",
     "engines": {
-        "vscode": "^1.22.0"
+        "vscode": "^1.23.0"
     },
     "categories": [
         "Other"
--- a/extension/package.json
+++ b/extension/package.json
@@ -42,7 +42,7 @@
             "./server/com.microsoft.java.test.plugin-0.6.0.jar"
         ],
         "views": {
-            "explorer": [
+            "test": [
                 {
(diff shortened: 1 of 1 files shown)


extension/package.json
```diff
@@ -42,7 +42,7 @@
             "./server/com.microsoft.java.test.plugin-0.6.0.jar"
         ],
         "views": {
-            "explorer": [
+            "test": [
```
yaohaizh: Do we need update the VSCode engine version for this?

sandy081: @yaohaizh Yes good point. Will update that.

ansyral:
@sandy081 thanks for your contribution, could you help update https://github.com/Microsoft/vscode-java-test/blob/master/.travis.yml#L11?

sandy081:
@ansyral Does the value needs to be set to 1.23?

ansyral:
@sandy081  yes, we have to update it, otherwise the test won't pass.

sandy081:
@ansyral I was on vacation and hence could not get back on to this.

Created another PR for the travis change -
https://github.com/Microsoft/vscode-java-test/pull/196

Issue #198, opened 2018-05-18
Restart is necessary after changing dependencies with Gradle


mouadino:
**VS Code:** 1.23.1
**Java Test Runner:** 0.6.1

Each time I add a new dependency to `build.gradle`, tests fail in case they need the specific class (e.g. JDBC driver) while running `gradle test` from terminal works, and it seems (happens already 2 times) that it only works after I restart vs code.

ansyral:
@mouadino , thanks for the report. Seems class path are not updated automatically when updating `build.gradle`. Will fix it.

PR #211, opened 2018-06-20
update classpath when user updates pom.xml or build.gradle

### LINKED ISSUE #205: Restart is necessary after changing dependencies with Maven
**VS Code**: 1.23.1
**Java Test Runner**: 0.6.1

Each time I add a new dependency to `pom.xml`, tests fail because they need some specific classes from the new dependency (throws error `NoClassDefFoundErrors`). Running `$ mvn test` from terminal works fine.

Test Runner only works again after I restart VSCode.

Possibly related issues: #198 , #138

ansyral:

to resolve #205 #198 

Currently there is no event fired when language server updates the project, so I have to add a timeout, which is 10 seconds

### DIFF
--- a/extension/src/extension.ts
+++ b/extension/src/extension.ts
@@ -80,6 +80,13 @@ export async function activate(context: ExtensionContext) {
         const uri = document.uri;
         testResourceManager.setDirty(uri);
         onDidChange.fire();
+        const fileName = path.win32.basename(uri.fsPath).toLowerCase();
+        if (fileName === 'pom.xml' || fileName === 'build.gradle') {
+            // add a timeout as there is no event to listen when JDTLS updates the project.
+            setTimeout(async () => {
+                await classPathManager.refresh();
+            }, 10000);
+        }
     });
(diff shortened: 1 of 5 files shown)


extension/src/extension.ts
```diff
@@ -80,6 +80,13 @@ export async function activate(context: ExtensionContext) {
         const uri = document.uri;
         testResourceManager.setDirty(uri);
         onDidChange.fire();
+        const fileName = path.basename(uri.fsPath).toLowerCase();
+        if (fileName === 'pom.xml' || fileName === 'build.gradle') {
```
yaohaizh: Is this reliable way to monitor the language server event? We should better off to use the language server side build_complete for this. THe server can disable auto build. I

ansyral: I think this pr should be a walkaround. In the long term, if language server doesn't export such event, we might need to open a feature request or pr to them to add this as I think it is also useful for other extensions that need such language server support.

Issue #200, opened 2018-05-21
not possbile to set environment variables


thwfreak:
there is no option in launch.test.json to set environment variables as in launch.json.

my test cases are configured via these env variables and so it is not possible for me to run these tests in vscode. 

Any plans or ideas for supporting env variables for the test environment?

ansyral:
@thwfreak  thanks for the suggestion, yes, we would add support for env variables later.

PR #215, opened 2018-06-20
support environment variables for test configuration

ansyral:

#200

### DIFF
--- a/extension/src/Models/testConfig.ts
+++ b/extension/src/Models/testConfig.ts
@@ -17,5 +17,6 @@ export type RunConfigItem = {
     workingDirectory: string;
     args: any[];
     vmargs: any[];
+    env: {[key: string]: string; };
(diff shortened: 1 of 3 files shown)


extension/src/Models/testConfig.ts
```diff
@@ -17,5 +17,6 @@ export type RunConfigItem = {
     workingDirectory: string;
     args: any[];
     vmargs: any[];
+    env: {[key: string]: string; };
```
yaohaizh: If the value is not required, should add ? for declaration

thwfreak:
I think this new configuration should be added to vscode-java-test/extension/launch.test.schema.json

ansyral:
@thwfreak nice catch! Would you like to open a pull request to help fix this?

thwfreak:
@ansyral  done with #225 :wink:

PR #219, opened 2018-06-25
change report refresh behavior when test files are changed but the test hasn't been rerun

ansyral:
previous: show error report
now: error snippets are prepended in previous report so user could still view previous report before rerunning the test.


ansyral:
It would show as below now:

image

ansyral:
Changed the color a bit:

image

PR #225, opened 2018-07-03
added env variables to launch.test.schema.json

thwfreak:
provide launch.test.json schema description for #215


ansyral:
@thwfreak Thanks!

Issue #240, opened 2018-07-09
should invoke a runner per project


ansyral:
scenario: if the user triggers a list of tests from multiple projects(say, trigger from test explorer), then should invoke a runner for each project as the different projects might contain conflict dependency and if we use a runner to run them all together, the class path might be resolved wrongly.

Issue #236, opened 2018-07-06
Test explorer sometimes detect wrong tests when there have test functions with same name in different class


JanusWu:
**To Reproduce**
1. Open project _junit5-samples_ in vscode
2. Open the Test Explorer, wait for the finish of test detecting
3. Open all java files one by one to let all tests show in the Test Explorer
4. There would have 3 _addsTwoNumbers_ test under class _CalculatorTetsts_
5. Check the file paths of three _addsTwoNumbers_ functions, you will see two of the functions have the same file path. (to get the file path, you can click on one of the function, move your mouse to hover over the opened java file title)

![image](https://user-images.githubusercontent.com/17060520/42371287-9692adfe-8141-11e8-99ba-5e9bb46ff0a0.png)

![image](https://user-images.githubusercontent.com/17060520/42371323-b3e98210-8141-11e8-8854-eff5481c33cd.png)

I checked all the _addsTwoNumbers_ functions (or CalculatorTests.java file) under _junit5-samples_ folder, and found one under project _junit5-jupiter-starter-gradle_, one under _junit5-jupiter-starter-ant_ and one under _junit5-jupiter-starter-maven_.
But in Test Explorer, one of the functions is under _junit5-jupiter-starter-gradle_, two of them are under  _junit5-jupiter-starter-maven_.

ansyral:
@JanusWu thanks for the report, this is a known issue. Currently, runner uses fully qualified name of a test method as the unique identifier, which is not enough for the case above.

ansyral:
root cause is same as #212

bsaby:
Build: 338
The current behavior is that if there are some test files with the same name, the Test Explorer will only display one of the files. Other identical name files also have no code lens.

Test explorer:
![image](https://user-images.githubusercontent.com/22470312/48337238-5d0a2500-e69d-11e8-9bc1-1a97aaaf65b7.png)
Other identical name files: 
![image](https://user-images.githubusercontent.com/22470312/48337285-7f03a780-e69d-11e8-9226-58063df06b0f.png)

jdneo:
@bsaby This is as expected.

Because the same named files contain the classes which have the same fully qualified names. They are introduced by different building tools. But the underlying Java Element is the same.

Issue #212, opened 2018-06-20
Can’t open test report by clicking status bar item


bsaby:
Repro steps:
1. clone code from https://github.com/spring-cloud/spring-cloud-function/tree/master/spring-cloud-function-samples
2. Run all tests

Results:
Can't open test report by clicking status bar item
![image](https://user-images.githubusercontent.com/22470312/41645932-0aa2a590-74a5-11e8-8978-cefab17c5e54.png)

bsaby:
![image](https://user-images.githubusercontent.com/22470312/41710828-384fe17c-7579-11e8-81b1-8c956b10c220.png)

ansyral:
tests are now identified with its qualified name, which might be duplicate if two projects use same package name and method name, and thus leading to the weird behavior

PR #246, opened 2018-07-11
add project name into test suite

ansyral:

fully qualified name of a test method isn't enough for the scenario that multi-projects are in a workspace folder


ansyral:
#212 #236 #240

PR #254, opened 2018-07-19
fix bug according to bi

ansyral:

### DIFF
--- a/extension/src/testConfigManager.ts
+++ b/extension/src/testConfigManager.ts
@@ -51,7 +51,7 @@ export class TestConfigManager {
             throw new Error('Not supported without a folder!');
         }
         const editor = window.activeTextEditor;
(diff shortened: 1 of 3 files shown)


extension/src/testConfigManager.ts
```diff
@@ -51,7 +51,7 @@ export class TestConfigManager {
             throw new Error('Not supported without a folder!');
         }
         const editor = window.activeTextEditor;
-        let folder = workspace.getWorkspaceFolder(editor.document.uri);
+        let folder = editor && workspace.getWorkspaceFolder(editor.document.uri);
```
yaohaizh: editor & editor.document?

ansyral: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_Operators

PR #265, opened 2018-08-06
not use gson to serialize test output

ansyral:


yaohaizh:
Please be careful about your commit message. See the information @akaroml  provide yesterday for commits.

com.microsoft.java.test.runner/src/main/java/com/microsoft/java/test/runner/Pair.java
```diff
@@ -0,0 +1,11 @@
+package com.microsoft.java.test.runner;
+
+public class Pair {
```
yaohaizh: `final` class?

yaohaizh: Could use this https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/tuple/package-summary.html class since our binary is already depends on commons package

com.microsoft.java.test.runner/src/main/java/com/microsoft/java/test/runner/TestMessageItem.java
```diff
@@ -0,0 +1,38 @@
...
+
+import java.io.PrintWriter;
+import java.io.StringWriter;
+import java.util.Arrays;
+import java.util.List;
+
+public class TestMessageItem {
+    String name;
+
+    List<Pair> attributes;
```
yaohaizh: Why not use hashmap?

com.microsoft.java.test.runner/src/main/java/com/microsoft/java/test/runner/TestOutputStream.java
```diff
@@ -0,0 +1,107 @@
...
+
+    @Override
+    public void flush() {
+        this.out.flush();
+        this.err.flush();
+    }
+
+    private static String toJson(TestMessageItem item) {
+        StringBuilder builder = new StringBuilder("@@<TestRunner-{\"name\":");
+        builder.append('"').append(item.name).append('"');
```
yaohaizh: What is the issue for using GSON?

com.microsoft.java.test.runner/src/main/java/com/microsoft/java/test/runner/TestMessageItem.java
ansyral: if using a hashmap, when creating an instance of TestMessageItem, it would introduce a lot of code blocks to create a hashmap and then set the key value pair. So I just use a list instead.

com.microsoft.java.test.runner/src/main/java/com/microsoft/java/test/runner/TestOutputStream.java
ansyral: GSON is an external dependency, I have to pack the dependency into the jar file and it would conflict for some projects. I'm still investigating the root cause, but I'd like to change it to previous implementation before the root cause is clear.

PR #277, opened 2018-08-28
 fix bug: test storage won't be updated when a test file is deleted

ansyral:


extension/src/extension.ts
```diff
@@ -76,6 +66,14 @@ export async function activate(context: ExtensionContext) {
         // onDidChange.fire();
     });
 
+    workspace.onDidCloseTextDocument((document) => {
+        const uri = document.uri;
+        // if a test file is removed, should update test storage.
+        if (!fs.existsSync(uri.fsPath)) {
+            testResourceManager.storeTests(uri, undefined);
```
jdneo: remove tests in the method named `storeTests` is confusing. Can we make another method named something like `removeTests()`?

ansyral: make sense

extension/src/extension.ts
```diff
@@ -59,26 +59,14 @@ export async function activate(context: ExtensionContext) {
...
-        const uri = document.uri;
-        // if a test file is removed, should update test storage.
-        if (!fs.existsSync(uri.fsPath)) {
-            testResourceManager.removeTests(uri);
-        }
-    });
-
-    workspace.onDidSaveTextDocument((document) => {
-        const uri = document.uri;
+    const watcher = workspace.createFileSystemWatcher('**/*.{[jJ][aA][vV][aA]}');
```
jdneo: Looks like we can push the `watcher` into `context.subscription`?

https://github.com/Microsoft/vscode/blob/80d6fa0b8eb5f4c0959b9e2925ec7fd28692880f/src/vs/vscode.d.ts#L1490
```javascript
export interface FileSystemWatcher extends Disposable
```

ansyral: yes, we should!

PR #278, opened 2018-08-29
Add more checks in tsconfig.json

jdneo:


extension/tsconfig.json
```diff
@@ -7,7 +7,11 @@
             "es6"
         ],
         "sourceMap": true,
-        "rootDir": "."
```
yaohaizh: Thoughts: 

We have lots of ts project or java project. Could we share the configs for the team wide project? Just like airbnb's js configuration.

Issue #259, opened 2018-07-26
Load test cases only when necessary (Delay load)


akaroml:
Loading test cases occupies considerable amount of time of project load. Check the performance test result below:

<img width="774" alt="screen shot 2018-07-26 at 9 42 08 pm" src="https://user-images.githubusercontent.com/16755094/43266113-16b326ae-911d-11e8-8cef-303b9848aaa0.png">

There is a significant gap between loading a project with or without the test runner.

Please delay load test cases to the moment when users show the intentions to use them. Here are some possible scenarios (but not limited):
- When users toggle the visibility of Test Explorer
- When users open a Java file with tests and code lens needs to be visible
- When users run test cases by triggering some command

As a result, we reduce the chance of loading all test cases along with project load to improve experience.

ansyral:
Currently we fetch all the tests and store them into the storage when the extension is activated. So to delay fetching tests until the user accesses Test Explorer , I plan to do the following changes:
1. Not load all the tests from the beginning(`TestResourceManager.refresh()`).
2. For Test Explorer, fetch the children and update test storage when `getChildren()` is called by vscode.
3. Add incremental refresh function for test explorer.
if a class is updated(add/remove/update test methods), reload its parent, this class and its children.

ansyral:
Found one bug test: storage isn't updated when a test file is removed. Opened a pr #277

akaroml:
Appreciate the effort of breaking down the problem. The logic of resolving test cases level by level looks fine to me. My thought is regarding

> I don't want the packages that doesn't contain tests to show in this tree. If we resolve layer by layer, then when resolving the package layer, we call language server to get all the packages defined in this workspace folder. We have no idea whether this package contains test unless we fetch its children.

For this part, I think it's fine to show packages at the top level when they contain no test cases. Hiding the empty ones cause extra computation that I don't think worthy. At the same time, showing all the packages is providing a view that is symmetric to the package truth. So my suggestion would be keeping all the packages at the top level, load their children only when user expand them.

jdneo:
### Update:

Current behavior of the explorer: 
![lazyload](https://user-images.githubusercontent.com/6193897/45273330-cc6c6880-b4e4-11e8-840e-bd2c7828129a.gif)

Some todo tasks:
- Clean up the code which is used to load all the tests at once
- Run regression tests
- Fix regression bugs

jdneo:
Closing since we have changed to lazy-load in 0.11.0

PR #290, opened 2018-09-10
Implement test explorer lazyload - TypeScript part

jdneo:
This PR is to reimplement the Test explorer: using lazy load strategy. #259 

GIF: 
lazyload

Future tasks:
- Currently, we wrap the `TestSuite` into the `SearchResult`, we should consider simplifying the protocol in the future.


extension/src/Explorer/testExplorer.ts
```diff
@@ -26,20 +27,47 @@ export class TestExplorer implements TreeDataProvider<TestTreeNode> {
...
+                    result.displayName,
+                    result.suite.test,
+                    result.nodeType,
+                    result.suite.uri,
+                    result.suite.range,
+                    element,
+                ));
+                resolve(element.children);
+            });
+        } else {
```
akaroml: Wrap less code in control blocks. There are 3 branches and the third one can be at the second place. So the current second one can be out of the control block. Code is generally easier to read when there are less control blocks.
```ts
if (!element) {
  ...
} else if (element.children) {
  ...
}

return new Promise(async...);
```

PR #303, opened 2018-09-17
Always asking LS to get the tests information when providing the Code Lens

jdneo:
When the user is editing the test files, only LS will know the range update for each test entry. So always ask LS for test information rather than ask the local cache.


ansyral:
what about the case that user opened a test file, close it and reopen it again(or focus on a test file, deactivate and activate again)?

jdneo:
It is true that the cache can help when users open -> close -> open a file.

While this case is not happening quite often. When users are writing test cases. They may focus on the file and keep typing. Which makes the cache almost never be hit. That's why I think always fetching from the LS is better.

Another concern is that the code lens provider is not triggered in our control. So when the user is editing the file. the extension will trigger the `provideCodeLenses()` and set the cache dirty in the meantime. Which may cause the cache is set to dirty after `provideCodeLenses()` finished.

akaroml:
Correctness > optimization

Issue #310, opened 2018-09-20
Sort packages alphabetically in Test Explorer


jbelford:
*Feature request*

PR #312, opened 2018-09-21
Sort Items in Test Explorer

jbelford:
this is a quick change for the item i mentioned in #310


jdneo:
It works as expected.

image

@jbelford Thank you for the contribution!

PR #327, opened 2018-09-28
Show the text document when select tree node

jdneo:


extension/src/commands/explorerSelect.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+import { TextDocument, Uri, window, workspace } from 'vscode';
+import { TestTreeNode } from '../explorer/TestTreeNode';
+
+export async function select(node: TestTreeNode): Promise<void> {
+    const document: TextDocument = await workspace.openTextDocument(Uri.file(node.fsPath));
+    await window.showTextDocument(document, {preserveFocus: true, selection: node.range});
```
yaohaizh: 1. Where is the return value
2. This function is mainly to open a file, but not select a file. 
3. If need select a node, should use the TreeView.reveal to select a node.

extension/src/constants/commands.ts
```diff
@@ -8,3 +8,7 @@ export enum JavaLanguageServerCommands {
 export enum JavaTestRunnerDelegateCommands {
     SearchTestItems = 'vscode.java.test.search.items',
 }
+
```
yaohaizh: Not sure why use enum value here

PR #328, opened 2018-09-29
Add code for searching the code lenses - Java

jdneo:
codelens

### DIFF
--- a/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/searcher/TestFrameworkSearcher.java
+++ b/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/searcher/TestFrameworkSearcher.java
@@ -14,15 +14,10 @@ package com.microsoft.java.test.plugin.searcher;
 import com.microsoft.java.test.plugin.model.TestKind;
 
 import org.eclipse.jdt.core.IMethod;
-import org.eclipse.jdt.core.search.SearchPattern;
 
 public interface TestFrameworkSearcher {
 
-    SearchPattern getSearchPattern();
-
     TestKind getTestKind();
-
-    String getTestMethodAnnotation();
 
     boolean isTestMethod(IMethod method);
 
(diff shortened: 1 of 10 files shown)


com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestSearchUtils.java
```diff
@@ -76,6 +87,46 @@
...
 
+    public static List<TestItem> searchCodeLens(List<Object> arguments, IProgressMonitor monitor)
+            throws OperationCanceledException, InterruptedException, JavaModelException {
+        final List<TestItem> resultList = new ArrayList<>();
+        if (arguments == null || arguments.size() == 0) {
+            return resultList;
+        }
+        final String uri = (String) arguments.get(0);
+        final ICompilationUnit unit = JDTUtils.resolveCompilationUnit(uri);
+        if (unit == null || !unit.getResource().exists() || !isInTestScope(unit) || monitor.isCanceled()) {
```
Eskibear: Will `unit.getResource()` be null?

com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/searcher/TestFrameworkSearcher.java
```diff
@@ -13,14 +13,17 @@
 
 import com.microsoft.java.test.plugin.model.TestKind;
 
+import org.eclipse.jdt.core.IMethod;
 import org.eclipse.jdt.core.search.SearchPattern;
 
-public abstract class JUnitTestSearcher {
+public interface TestFrameworkSearcher {
```
akaroml: This is a good start. But I don't see getSearchPattern and getTestMethodAnnotation called anywhere. Do we really need those two?

jdneo: Removed those unused methods.

PR #340, opened 2018-10-10
Remove the legacy code for search the code lens

jdneo:
These codes are the legacy codes used to search the code lens.\

Since we have implemented within new logic, these codes can be removed.


Eskibear:
Also remove `vscode.java.test.fetch` from plugin.xml.

PR #353, opened 2018-10-24
Junit 5 support

jdneo:


com.microsoft.java.test.runner/src/main/java/com/microsoft/java/test/runner/junit5/CustomContextClassLoaderExecutor.java
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright 2015-2018 the original author or authors.
```
yaohaizh: Need refine the license header. Add Microsoft or other specific

PR #360, opened 2018-10-31
Add config manager to manage the test config for each run/debug

jdneo:


extension/src/runConfigs.ts
```diff
@@ -0,0 +1,22 @@
...
+    projectName: string;
+    workingDirectory: string;
+    args: any[];
+    vmargs: any[];
+    env: { [key: string]: string; };
+    preLaunchTask: string;
+}
+
+export interface IExecutionConfigGroup {
+    default: string;
```
yaohaizh: What is this value for?

jdneo: `default` is the name of the default `IExecutionConfig` in `IExecutionConfigGroup`

PR #361, opened 2018-10-31
Add classpath util methods to write classpaths into a jar

jdneo:


extension/src/utils/classpathUtils.ts
```diff
@@ -0,0 +1,44 @@
...
+
+import * as archiver from 'archiver';
+import * as fileUrl from 'file-url';
+import * as fse from 'fs-extra';
+import * as os from 'os';
+import * as path from 'path';
+import { MAX_CLASS_PATH_LENGTH } from '../constants/configs';
+
+export async function getClassPathString(classpaths: string[], storagePath: string): Promise<string> {
+    const separator: string = process.platform === 'darwin' || process.platform === 'linux' ? ':' : ';';
```
yaohaizh: Should factor out code to handle platform related issues

PR #362, opened 2018-10-31
Add model definition for runners

jdneo:


extension/src/protocols.ts
```diff
@@ -15,10 +18,8 @@ export interface ITestItem {
     level: TestLevel;
 }
 
-export interface ISearchChildrenNodeRequest {
+export interface ISearchChildrenNodeRequest extends ITestItemBase {
```
yaohaizh: Naming confusion here

extension/src/runners/models.ts
```diff
@@ -0,0 +1,22 @@
...
+export interface ITestResultDetails {
+    status?: TestStatus;
+    details?: string;
+    message?: string;
+    duration?: string;
+    summary?: string;
+}
+
+export enum TestStatus {
+    Pass = 'Pass',
```
yaohaizh: Pass, Fail Skip or Passed, Failed, Skipped

PR #364, opened 2018-10-31
Add base runner class

jdneo:


extension/src/utils/cpUtils.ts
```diff
@@ -0,0 +1,19 @@
...
+
+import * as cp from 'child_process';
+import * as kill from 'tree-kill';
+
+export function killProcess(process: cp.ChildProcess): Promise<void> {
+    return new Promise<void>((resolve: () => void, reject: (err: Error) => void): void => {
+        if (process) {
+            kill(process.pid, 'SIGTERM', (error: Error | undefined) => {
+                if (error) {
+                    reject(error);
```
Eskibear: The code here doesn't have any problem, but it's better to return early after reject. 

See a post [here](https://stackoverflow.com/questions/32536049/do-i-need-to-return-after-early-resolve-reject)

extension/src/runners/baseRunner/BaseRunner.ts
```diff
@@ -0,0 +1,155 @@
...
+    protected isCanceled: boolean;
+
+    constructor(
+        protected javaHome: string,
+        protected storagePath: string) {}
+
+    public abstract getTestResultAnalyzer(): BaseRunnerResultAnalyzer;
+
+    public get serverHome(): string {
+        return path.join(__dirname, '..', '..', '..', '..', 'server');
```
yaohaizh: Use extensionPath for this value?

extension/src/runners/baseRunner/BaseRunnerResultAnalyzer.ts
```diff
@@ -0,0 +1,73 @@
...
+                try {
+                    this.processData(match[1]);
+                } catch (ex) {
+                    // Swallow
+                }
+            }
+        } while (match);
+    }
+
+    public analyzeError(_error: string): void {
```
yaohaizh: Why start with underscore?

jdneo: Because it's not used param for now. This is to dismiss the `noUnusedParameters` check

PR #365, opened 2018-11-01
Add Junit4 runner

jdneo:


extension/src/runners/junit4Runner/Junit4RunnerResultAnalyzer.ts
```diff
@@ -0,0 +1,62 @@
...
+import { ITestResultDetails, TestStatus } from '../models';
+
+const TEST_START: string = 'testStarted';
+const TEST_FAIL: string = 'testFailed';
+const TEST_FINISH: string = 'testFinished';
+
+export class JUnit4RunnerResultAnalyzer extends BaseRunnerResultAnalyzer {
+
+    protected processData(data: string): void {
+        let res: ITestResultDetails | undefined;
```
yaohaizh: Where is the place the res be get but not set?

PR #373, opened 2018-11-01
Add way to disable test and debug code lens provider

ThadHouse:
There are cases where the tests by default do not work correctly, and
the launch.test.json is not configurable enough to work trivially with
some of my projects (mainly JNI projects). I have had a large amount of
users get confused why these buttons don't work, and attempts at using
the launch.test.json file don't work either. I provided this same PR to
the debug extension as well.

### DIFF
--- a/extension/README.md
+++ b/extension/README.md
@@ -6,7 +6,7 @@
 
 A lightweight extension to run and debug Java test cases in Visual Studio Code. It works with [Language Support for Java by Red Hat](https://marketplace.visualstudio.com/items?itemName=redhat.java) and [Debugger for Java](https://marketplace.visualstudio.com/items?itemName=vscjava.vscode-java-debug) to provide the following features:
 
-- Support JUnit (v4.8.0+) test cases 
+- Support JUnit (v4.8.0+) test cases
 - Run test cases
 - Debug test cases
 - View test logs
@@ -28,6 +28,7 @@ A lightweight extension to run and debug Java test cases in Visual Studio Code. 
 ## Supported VSCode settings
 
 - `java.test.report.position`: Specify where to show the test report. Supported values are `sideView`, `currentView`. The default value is `sideView`.
+- `java.test.settings.enableTestCodeLens`: enable the code lens provider for the test and debug buttons over test entry points, defaults to `true`.
 
 ## Requirements
 
--- a/extension/README.md
+++ b/extension/README.md
@@ -28,7 +28,7 @@ A lightweight extension to run and debug Java test cases in Visual Studio Code. 
 ## Supported VSCode settings
 
 - `java.test.report.position`: Specify where to show the test report. Supported values are `sideView`, `currentView`. The default value is `sideView`.
-- `java.test.settings.enableTestCodeLens`: enable the code lens provider for the test and debug buttons over test entry points, defaults to `true`.
+- `java.test.settings.enableTestCodeLens`: Enable the code lens provider for the test and debug buttons over test entry points, defaults to `true`.
 
 ## Requirements
 
--- a/extension/package.json
+++ b/extension/package.json
@@ -190,6 +190,11 @@
                     "default": "sideView",
                     "description": "Specify where to show the test report",
                     "scope": "window"
+                },
+                "java.test.settings.enableTestCodeLens": {
+                    "type": "boolean",
+                    "description": "Enable the test code lens providers over test methods.",
+                    "default": true
                 }
             }
         }
--- a/extension/package.json
+++ b/extension/package.json
@@ -193,7 +193,7 @@
                 },
                 "java.test.settings.enableTestCodeLens": {
                     "type": "boolean",
-                    "description": "Enable the test code lens providers over test methods.",
+                    "description": "Enable the test code lens providers over test methods and classes.",
                     "default": true
                 }
             }
--- a/extension/src/codeLensContainer.ts
+++ b/extension/src/codeLensContainer.ts
@@ -1,0 +1,44 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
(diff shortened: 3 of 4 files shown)


jdneo:
@ThadHouse Thank you for the contribution. Since this change could have an impact on the current user experience. Could you please also create a tracking issue for this PR? Meanwhile, we can wait a while to see if any other users are requesting for the same feature. 😃

ThadHouse:
Done. Note the behavior is not changed by default. The setting is defaulted to true, so it doesn't change anything without the user explicitly changing their settings.

How long would I have to wait? Our product is releasing at the beginning of January, and we would like to have pretty much final builds by the beginning of December at the latest. This is not a breaking change, so it shouldn't take too long hopefully.

ThadHouse:
For providing a sample, I need to set the working directory of the executor to a directory generated by gradle during its test configuration. So every time a test is requested, I need to be able to run a gradle build, receive input from that gradle build and set the working directory of the test to a part of that build.

The other option would be to have a way for this extension to smartly hook into gradle to get test parameters, but that's a much more difficult thing to do (I've looked, and its not super easy).

ThadHouse:
[FinalJavaTest2.zip](https://github.com/Microsoft/vscode-java-test/files/2540818/FinalJavaTest2.zip)

@jdneo Here is a sample that reproduces some of the issues I see. Note that in my use case the build folder is not always deterministic (in this example it is, but in my project its not), so the working directory of the tests might be different for every build.

jdneo:
Thanks for sharing, we'll look at it.

ThadHouse:
@jdneo Is there any blocker on this other then the conflicting files? I can fix those. This is still a blocking issue for me.

jdneo:
Hi @ThadHouse, 

Sorry that we did some refactoring these days causing the conflicts in the PR. It will be great if you would like to resolve this.

We decoupled several components now, so I think you can just check the configuration and determine if the CodeLensProvider needs to be registered in `extension.ts`. What do you think?

extension/src/Constants/constants.ts
```diff
@@ -8,3 +8,5 @@ export const TELEMETRY_EXCEPTION_SCOPE = 'exception';
 export const TELEMETRY_INFO_SCOPE = 'info';
 export const TELEMETRY_WARN_SCOPE = 'warn';
 export const TEST_EXPLORER_VIEW_ID = 'testExplorer';
+export const JAVA_TEST_SETTINGS_CONFIGURATION = 'java.debug.settings';
```
jdneo: I guess you want to say `java.test` instead of `java.debug`?

ThadHouse:
@jdneo I was able to get this updated. It should be good to go now.

extension/README.md
```diff
@@ -28,6 +28,7 @@ A lightweight extension to run and debug Java test cases in Visual Studio Code.
 ## Supported VSCode settings
 
 - `java.test.report.position`: Specify where to show the test report. Supported values are `sideView`, `currentView`. The default value is `sideView`.
+- `java.test.settings.enableTestCodeLens`: enable the code lens provider for the test and debug buttons over test entry points, defaults to `true`.
```
jdneo: nit: `enable` -> Enable

extension/package.json
```diff
@@ -190,6 +190,11 @@
                     "default": "sideView",
                     "description": "Specify where to show the test report",
                     "scope": "window"
+                },
+                "java.test.settings.enableTestCodeLens": {
+                    "type": "boolean",
+                    "description": "Enable the test code lens providers over test methods.",
```
jdneo: nit: methods and classes

ThadHouse:
You can still access the TestConfigurationProvider like before, as there. Is still only a singleton one. The container is needed to make it possible to deactivate and reactive the provider as the user changes the setting.

It would be possible to have a whole configuration manager, but that's a little overkill. Most extensions I've looked at just grab settings when they need them, and don't actually depend on the configuration changed event. This only needs the configuration changed event, as it activates and deactivates the provider dynamically. Most settings don't need this functionality.

I'll fix the other things in a bit.

ThadHouse:
I fixed all the review comments. This should be good now.

extension/src/codeLensContainer.ts
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+import { ConfigurationChangeEvent, Disposable, languages, workspace, WorkspaceConfiguration } from 'vscode';
+import { testCodeLensProvider } from './codeLensProvider';
+
+const JAVA_TEST_SETTINGS_CONFIGURATION: string = 'java.test.settings';
```
jdneo: Might be better to put them into https://github.com/Microsoft/vscode-java-test/blob/master/extension/src/constants/configs.ts

extension/src/codeLensContainer.ts
```diff
@@ -0,0 +1,44 @@
...
+// Licensed under the MIT license.
+
+import { ConfigurationChangeEvent, Disposable, languages, workspace, WorkspaceConfiguration } from 'vscode';
+import { testCodeLensProvider } from './codeLensProvider';
+
+const JAVA_TEST_SETTINGS_CONFIGURATION: string = 'java.test.settings';
+const ENABLE_CODE_LENS_VARIABLE: string = 'enableTestCodeLens';
+const LANGUAGE: string = 'java';
+
+class TestCodeLensContainer implements Disposable {
```
jdneo: Change the class name to `TestCodeLensManager`?

extension/src/codeLensContainer.ts
```diff
@@ -0,0 +1,44 @@
...
+  private lensProvider: Disposable | undefined;
+  private configurationEvent: Disposable;
+
+  constructor() {
```
jdneo: Though register the listeners in the constructor can make it work here. I still prefer to using a method called `initialize()` and call it explicitly instead of put listener registration in the constructor.

jdneo:
Addressed in #734

PR #375, opened 2018-11-02
Add executor for all the runners

jdneo:


extension/src/runners/runnerExecutor.ts
```diff
@@ -0,0 +1,154 @@
...
+import { TestNGRunner } from './testngRunner/TestNGRunner';
+
+export class RunnerExecutor {
+    private readonly javaHome: string;
+    private readonly context: ExtensionContext;
+    private isRunning: boolean;
+    private preLaunchTask: cp.ChildProcess | undefined;
+    private runnerMap: Map<ITestRunner, ITestItem[]> | undefined;
+
+    constructor(javaHome: string, context: ExtensionContext) {
```
yaohaizh: Use _javaHome, _context to simplify the private local variable declaration

```js
  constructor(private _javaHome: string, private _context: ExtensionContext) {
```

extension/src/runners/runnerExecutor.ts
```diff
@@ -0,0 +1,154 @@
...
+            window.showErrorMessage(`${error}`);
+        } finally {
+            await this.cleanUp();
+        }
+    }
+
+    private async cleanUp(): Promise<void> {
+        try {
+            if (this.preLaunchTask) {
+                await killProcess(this.preLaunchTask);
```
yaohaizh: What happens if the preLaunchTask is already closed?

jdneo: Changed the logic: swallow the error when we kill the process

extension/src/runners/runnerExecutor.ts
```diff
@@ -0,0 +1,150 @@
...
+                this.preLaunchTask = undefined;
+            }
+
```
yaohaizh: Should swallow here? No matter what happens insider the preLaunchTask, need clean up the resource anyway.

PR #380, opened 2018-11-03
Add out put channel

jdneo:


extension/src/testOutputChannel.ts
```diff
@@ -0,0 +1,30 @@
...
+import { Disposable, OutputChannel, window } from 'vscode';
+
+class TestOutputChannel implements Disposable {
+    private readonly channel: OutputChannel = window.createOutputChannel('Java Test Runner');
+
+    public info(info: string): void {
+        this.appendLine(`[INFO] ${info}`);
+    }
+
+    public error(error: string): void {
```
Eskibear: Might take `error(message: string, error: Error)` into consideration for unified output format.

PR #382, opened 2018-11-05
Can cancel the running progress

jdneo:


extension/src/testStatusBarProvider.ts
```diff
@@ -22,6 +22,10 @@ class TestStatusBarProvider implements Disposable {
         this.update('$(sync~spin) Running tests...', 'white', 'View test output', JavaTestRunnerCommands.SHOW_TEST_OUTPUT);
     }
 
+    public showRunningFail(): void {
```
Eskibear: `showRunningFail` better naming.

PR #386, opened 2018-11-06
Can search all test items under a certain node

jdneo:
Can search all test items under a certain node - Java part


com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/model/TestLevel.java
```diff
@@ -15,17 +15,20 @@
 
 public enum TestLevel {
     @SerializedName("0")
-    FOLDER,
+    ROOT,
```
yaohaizh: Workspace root? How about multiple workspace?

jdneo: `ROOT` -> Workspace Root
`FOLDER` -> Workspace Folder

PR #388, opened 2018-11-06
Remove useless methods and better naming

jdneo:


com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestSearchUtils.java
```diff
@@ -184,6 +184,11 @@ public void acceptSearchMatch(SearchMatch match) throws CoreException {
         new SearchEngine().search(pattern, new SearchParticipant[] { SearchEngine.getDefaultSearchParticipant() },
                 scope, requestor, monitor);
 
+        for (final TestItem testClass : classMap.values()) {
```
Eskibear: Better to add comments for this logic.

PR #401, opened 2018-11-07
Do not show code lens when there is no result for the target test item

### LINKED ISSUE #393: Should not affect the state of other methods if running a single method
Run a method from code lens, after running job,  a new status icon is added as the third Code Lens action. But for other methods and this class, why the status icon change to the question mark?
image

Verify this behavior in marketplace version: 
After running a single method,  just show status icon for this method and class level.But thereis a regression issue #127 here: 
image

jdneo:
Fix #393

### DIFF
--- a/extension/src/testResultManager.ts
+++ b/extension/src/testResultManager.ts
@@ -28,8 +28,12 @@ class TestResultManager implements Disposable {
         return undefined;
     }
 
-    public hasResultWithUri(uriString: string): boolean {
-        return this.testResultMap.has(uriString);
+    public hasResultWithFsPath(fsPath: string): boolean {
+        return this.testResultMap.has(fsPath);
+    }
+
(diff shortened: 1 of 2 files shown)


extension/src/testResultManager.ts
```diff
@@ -28,8 +28,12 @@ class TestResultManager implements Disposable {
...
     }
 
-    public hasResultWithUri(uriString: string): boolean {
-        return this.testResultMap.has(uriString);
+    public hasResultWithFsPath(fsPath: string): boolean {
+        return this.testResultMap.has(fsPath);
+    }
+
+    public hasResultWithFsPathAndFullName(fsPath: string, testFullName: string): boolean {
+        return this.getResult(fsPath, testFullName) !== undefined;
```
yaohaizh: What about `null`?

PR #403, opened 2018-11-07
Swallow the error when killing the child process

### LINKED ISSUE #394: There are some error messages in test output when running or debugging each test
OS: Windows

Run or debug test form codelens
image

jdneo:
fix #394

### DIFF
--- a/extension/src/utils/cpUtils.ts
+++ b/extension/src/utils/cpUtils.ts
@@ -3,15 +3,11 @@
 
 import * as cp from 'child_process';
 import * as kill from 'tree-kill';
-import { testOutputChannel } from '../testOutputChannel';
 
 export function killProcess(process: cp.ChildProcess): Promise<void> {
     return new Promise<void>((resolve: () => void): void => {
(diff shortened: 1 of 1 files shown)


extension/src/utils/cpUtils.ts
```diff
@@ -3,15 +3,11 @@
...
 
 export function killProcess(process: cp.ChildProcess): Promise<void> {
     return new Promise<void>((resolve: () => void): void => {
-        if (process) {
-            kill(process.pid, 'SIGTERM', (error: Error | undefined) => {
-                if (error) {
-                    testOutputChannel.error('Failed to kill the process', error);
-                }
+        if (process && !process.killed) {
+            kill(process.pid, 'SIGTERM', () => {
```
yaohaizh: If not output, at least log the issue for program diagnostic. It seems test runner need logger?

jdneo: @yaohaizh How about adding logger in the next release (0.12.0)

PR #423, opened 2018-11-08
Check the kind is undefined or not

jdneo:
`!test.kind` will return true if the value of kind is `0`.

### DIFF
--- a/extension/src/runners/runnerExecutor.ts
+++ b/extension/src/runners/runnerExecutor.ts
@@ -103,7 +103,7 @@ class RunnerExecutor {
     private mapTestsByProjectAndKind(tests: ITestItem[]): Map<string, ITestItem[]> {
         const map: Map<string, ITestItem[]> = new Map<string, ITestItem[]>();
         for (const test of tests) {
-            if (!test.kind) {
+            if (test.kind === undefined || test.kind === null) {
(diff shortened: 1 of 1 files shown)


extension/src/runners/runnerExecutor.ts
```diff
@@ -103,7 +103,7 @@ class RunnerExecutor {
     private mapTestsByProjectAndKind(tests: ITestItem[]): Map<string, ITestItem[]> {
         const map: Map<string, ITestItem[]> = new Map<string, ITestItem[]>();
         for (const test of tests) {
-            if (!test.kind) {
+            if (test.kind === undefined || test.kind === null) {
```
Eskibear: Looks a bit wired. If normally `kind` should be a number, how about 
```ts
if (test.kind instanceof number)
```

extension/src/runners/runnerExecutor.ts
```diff
@@ -103,7 +103,7 @@ class RunnerExecutor {
     private mapTestsByProjectAndKind(tests: ITestItem[]): Map<string, ITestItem[]> {
         const map: Map<string, ITestItem[]> = new Map<string, ITestItem[]>();
         for (const test of tests) {
-            if (test.kind === undefined || test.kind === null) {
+            if (test.kind && test.kind in TestKind) {
```
Eskibear: wrong conditional

PR #432, opened 2018-11-09
Add a log in output channel when job canceled

jdneo:


extension/src/runners/runnerExecutor.ts
```diff
@@ -89,6 +89,10 @@ class RunnerExecutor {
                 this._runnerMap = undefined;
             }
             await Promise.all(promises);
+
+            if (isCancel) {
+                testOutputChannel.info('Test job canceled.');
```
andxu: is canceled

jdneo: Updated

andxu: Test job is canceled

Issue #443, opened 2018-11-22
'Run Test' in 0.11.0 stopped working.


NikolasKomonen:
Since the release of 0.11.0 when I call 'Run Test' I get 'Failed to run tests' and a message window saying 'Error'.

The repo I've been working with is here: https://github.com/angelozerr/lsp4xml
And tests have been working up until the new release.

![selection_009](https://user-images.githubusercontent.com/32624665/48917741-69893d00-ee56-11e8-8b4c-01a4691f84ab.png)

Try `lsp4xml/org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/services/XMLFormatterTest.java` to run a test.

jdneo:
@NikolasKomonen Thanks for reporting. I'll take a look.

jdneo:
Main cause is that in the code here: https://github.com/Microsoft/vscode-java-test/blob/master/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/RuntimeClassPathUtils.java#L42

We need to check the project nature as well.

jdneo:
Java Test Runner 0.11.1 has been released.

PR #444, opened 2018-11-23
Get the Java project directly instead of using the IProject when resolving the class path

jdneo:
resolve: #443


NikolasKomonen:
@jdneo Is it possible to get the VSIX that includes this fix? Or what would I run to build it locally? 

Thanks for the fix btw.

yaohaizh:
In order to build the fix, sync the code. on the root folder of the repo
```
cd ./extension
npm install 
npm install -g vsce
npm run buildserver
vsce package
```

PR #456, opened 2018-11-26
Fix the "Cannot read property 'indexOf' of undefined" bug when running JUnit 5 tests

### LINKED ISSUE #455: Failed to parse output data: Cannot read property 'indexOf' of undefined
I can see this log in the output channel when running JUnit 5 test cases.

The root cause is the regex here: https://github.com/Microsoft/vscode-java-test/blob/master/extension/src/runners/junit5Runner/JUnit5RunnerResultAnalyzer.ts#L47

jdneo:
Fix #455

### DIFF
--- a/extension/src/runners/junit5Runner/JUnit5RunnerResultAnalyzer.ts
+++ b/extension/src/runners/junit5Runner/JUnit5RunnerResultAnalyzer.ts
@@ -44,7 +44,7 @@ export class JUnit5RunnerResultAnalyzer extends BaseRunnerResultAnalyzer {
         if (!id) {
             return id;
         }
-        const regex: RegExp = /\[class:(.*?)\](?:\/\[method:(.*)\])?/gm;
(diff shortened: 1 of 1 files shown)


extension/src/runners/junit5Runner/JUnit5RunnerResultAnalyzer.ts
```diff
@@ -44,7 +44,7 @@ export class JUnit5RunnerResultAnalyzer extends BaseRunnerResultAnalyzer {
         if (!id) {
             return id;
         }
-        const regex: RegExp = /\[class:(.*?)\](?:\/\[method:(.*)\])?/gm;
+        const regex: RegExp = /\[class:(.*?)\]\/\[method:(.*)\]/gm;
         const match: RegExpExecArray | null = regex.exec(id);
         if (match) {
             if (match.length === 2) {
```
Eskibear: `match.length === 2` not possible for current regexp

PR #472, opened 2018-11-29
Remove the dependency of glob-promise

jdneo:


jdneo:
@Eskibear Changed as you advised

Issue #314, opened 2018-09-23
Test report bar should be either green or red, depending on the test status


fbricon:
By convention, in Java IDEs, running tests should result in either a green, or a red bar.
Right now, the status report sends confusing signals, show both green and red bars. 
In this example: the red bar is larger than the green one, even though nothing failed.
<img width="894" alt="screen shot 2018-09-23 at 7 22 36 pm" src="https://user-images.githubusercontent.com/148698/45934075-b9d34400-bf66-11e8-856b-a7b621420ca6.png">
Contrast that to a failing test:
<img width="899" alt="screen shot 2018-09-23 at 7 28 03 pm" src="https://user-images.githubusercontent.com/148698/45934081-dc655d00-bf66-11e8-9ea4-8737158c3804.png">

Even though I can read the text inside,  I find this visually very confusing. 

Primary bar should just be either green or red.

jdneo:
Updated the UI.

More details can be found in #475

jdneo:
Now the `fail` filter button will only appear when there are failed cases.

PR #475, opened 2018-11-29
Make the final result of the test report more clearly

jdneo:
Address the issue: #314 

Here are the screenshots for different scenarios:

### When having both passed cases and failed cases
image

### When only having passed cases
image

### When only having failed cases
image


Eskibear:
It looks good.

Issue #477, opened 2018-11-29
Got SecurityException when running JUnit 5 tests


jbelford:
I've had this issue for a while. Essentially, if I open up a project folder directly I can run JUnit5 tests. 

If I am in a multi-root workspace it will skip JUnit5 tests when I try to run them. JUnit4 tests seem to still work fine though.

Here's the log
```
.
[ERROR] @@<TestRunner-{"name":"error", "attributes":{"message":"Failed to run Junit tests", "details":
"java.lang.SecurityException: class \"org.junit.platform.engine.discovery.ModuleSelector\"'s signer information does not match signer information of other classes in the same package
  at java.lang.ClassLoader.checkCerts(ClassLoader.java:898)
  at java.lang.ClassLoader.preDefineClass(ClassLoader.java:668)
  at java.lang.ClassLoader.defineClass(ClassLoader.java:761)
  at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
  at java.net.URLClassLoader.defineClass(URLClassLoader.java:467)
  at java.net.URLClassLoader.access$100(URLClassLoader.java:73)
  at java.net.URLClassLoader$1.run(URLClassLoader.java:368)
  at java.net.URLClassLoader$1.run(URLClassLoader.java:362)
  at java.security.AccessController.doPrivileged(Native Method)
  at java.net.URLClassLoader.findClass(URLClassLoader.java:361)
  at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
  at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:349)
  at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
  at com.microsoft.java.test.runner.junit5.DiscoveryRequestCreator.createExplicitDiscoverySelectors(DiscoveryRequestCreator.java:83)
  at com.microsoft.java.test.runner.junit5.DiscoveryRequestCreator.createDiscoverySelectors(DiscoveryRequestCreator.java:61)
  at com.microsoft.java.test.runner.junit5.DiscoveryRequestCreator.toDiscoveryRequest(DiscoveryRequestCreator.java:44)
  at com.microsoft.java.test.runner.junit5.CustomizedConsoleTestExecutor.lambda$executeTests$0(CustomizedConsoleTestExecutor.java:41)
  at com.microsoft.java.test.runner.junit5.CustomContextClassLoaderExecutor.invoke(CustomContextClassLoaderExecutor.java:29)
[...]

PR #478, opened 2018-11-29
Remove selector for Module as it conflicts with multi-root workspaces (#477)

jbelford:
See #477 

I removed the module selector as it seems like it isn't really needed and causes problems when using multi-root workspaces.


jbelford:
I'm closing this PR as I believe @jdneo is onto a potential solution with signer information.

PR #512, opened 2018-12-07
Limit the log file's size

jdneo:


src/logger/logger.ts
```diff
@@ -15,7 +15,7 @@ class Logger implements Disposable {
         this.storagePath = storagePath;
         this.logger = winston.createLogger({
             transports: [
-                new (winston.transports.File)({level: 'info', filename: path.join(this.storagePath, LOG_FILE_NAME)}),
+                new (winston.transports.File)({level: 'info', filename: path.join(this.storagePath, LOG_FILE_NAME), maxsize: 5 * 1024 * 1024, maxFiles: 2, tailable: true}),
```
yaohaizh: Should avoid the hard code size

jdneo: Addressed in #513

Issue #517, opened 2018-12-11
Improve the test report


jdneo:
- [x] Add hover style
- [ ] Add arrow before each method line

PR #519, opened 2018-12-11
Change the row background color when hovering

jdneo:
Address #517 - Add hover style

table-hover

Updated:

table-hover

### DIFF
--- a/resources/templates/report_method_table.pug
+++ b/resources/templates/report_method_table.pug
@@ -10,7 +10,7 @@ mixin collapseMethodTable(children, type)
             - var methodIdx = 0
             each method in entry[1]
                 - methodIdx++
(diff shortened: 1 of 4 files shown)


resources/templates/report_method_table.pug
```diff
@@ -10,7 +10,7 @@ mixin collapseMethodTable(children, type)
             - var methodIdx = 0
             each method in entry[1]
                 - methodIdx++
-                div.row.accordion-toggle(data-toggle="collapse", data-target=`#${type}-${classIdx}-${methodIdx}`, style="cursor:pointer", title="View details")
+                div.row.row-custom.accordion-toggle(data-toggle="collapse", data-target=`#${type}-${classIdx}-${methodIdx}`, style="cursor:pointer", title="View details")
```
akaroml: What if you use the table-hover class?

PR #520, opened 2018-12-11
Refine the class name row in the report

jdneo:
Address #517 

image

image


resources/templates/scss/report.scss
```diff
@@ -15,5 +15,8 @@ $font-size-base: .85rem !default;
 pre.text { 
     color: var(--vscode-foreground);
 }
+.list-group-item-title {
```
akaroml: Can we change scss variable values instead?

jdneo: Updated

PR #521, opened 2018-12-12
Activate the extension when workspace contains pom.xml/build.gradle

### LINKED ISSUE #516: Activate the extension when the workspace contains pom.xml/build.gradle
Current the extension will be activated only if the extension command is triggered. Or user opened a java file.

We need to consider to change the activation event. For example https://github.com/redhat-developer/vscode-java/blob/master/package.json

Here are some changes that need to be done:
- [x] Activate the extension when the workspace contains pom.xml/build.gradle
- [ ] Activate the extension when test explorer gets focus

jdneo:
Resolve #516

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,8 @@
     ],
     "activationEvents": [
         "onLanguage:java",
+        "workspaceContains:pom.xml",
+        "workspaceContains:build.gradle",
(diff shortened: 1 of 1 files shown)


Eskibear:
I'm voting against this. It looks too aggressive to me.

Issue #437, opened 2018-11-15
Chinese Language Support


akaroml:
We have lots of Chinese developers. It would be great if we can provide a localized interface. This issue is to track the effort of building the localization story.
Supporting Chinese is a start. Once the solution is working, we can expand the support to other languages.

jdneo:
Closing this for now since we have [package.nls.zh.json](https://github.com/microsoft/vscode-java-test/blob/master/package.nls.zh.json) now.

Will reopen this it turns out that we have other more works to do.

PR #525, opened 2018-12-14
Support Chinese Locale for package.json

jdneo:
Part of the work for #437


package.nls.zh.json
```diff
@@ -0,0 +1,13 @@
+{
+    "description": "运行并调试 JUnit 或 TestNG 测试用例",
```
akaroml: No need to have leading and trailing spaces for English words.

PR #532, opened 2018-12-19
Using JDTUtils to expand the tree nodes instead of using SearchEngine

jdneo:


java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/CodeLensUtils.java
```diff
@@ -0,0 +1,81 @@
+/*******************************************************************************
+* Copyright (c) 2018 Microsoft Corporation and others.
+* All rights reserved. This program and the accompanying materials
+* are made available under the terms of the Eclipse Public License v1.0
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v10.html
```
yaohaizh: @akaroml How about the issue of EPL 2.0? Do we have any update on this part?

java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/CodeLensUtils.java
```diff
@@ -0,0 +1,81 @@
...
+import org.eclipse.jdt.ls.core.internal.handlers.DocumentLifeCycleHandler;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+@SuppressWarnings("restriction")
+public class CodeLensUtils {
```
yaohaizh: Decouple the underlying concept from the UI for flexibility.  

The CodeLens/TestExplorer are the UI elements. Here, we map the UI elements to the underlying utils, which will make the underlying fragile.  You should provide the API and let the UI decide how to call these APIs. Thus, you can decouple the UI from the underlying system.

java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestExplorerUtils.java
```diff
@@ -0,0 +1,245 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+@SuppressWarnings("restriction")
+public class TestExplorerUtils {
+
+    public static List<TestItem> searchTestItems(List<Object> arguments, IProgressMonitor monitor)
```
yaohaizh: Document the arguments? Maybe this is the issue of the API problem. I need redesign the execute command parameter restriction...

java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/CodeLensUtils.java
jdneo: Follow up:
- Remain the TestSearchUtils there
- Extract out the methods which are used to construct the TestItem into a new file: TestItemsUtils
- Extract out the methods which are used to deal with the test frameworks into a new file: TestFrameworkUtils.

java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestExplorerUtils.java
jdneo: Added JavaDoc to link to the related class

Issue #526, opened 2018-12-14
@ParameterizedTest skipped when invoked from method level 


jbelford:
@ParameterizedTest works when run from the class level but I run into errors when executed from the method level.

```
WARNING: TestEngine with ID 'junit-jupiter' failed to discover tests
org.junit.platform.commons.util.PreconditionViolationException: Could not find method with name [myArgumentMethod] in class [org.example.ExampleTest].
	at org.junit.platform.engine.discovery.MethodSelector.lambda$lazyLoadJavaMethod$2(MethodSelector.java:177)
	at java.util.Optional.orElseThrow(Optional.java:290)
	at org.junit.platform.engine.discovery.MethodSelector.lazyLoadJavaMethod(MethodSelector.java:175)
	at org.junit.platform.engine.discovery.MethodSelector.getJavaMethod(MethodSelector.java:142)
	at org.junit.jupiter.engine.discovery.DiscoverySelectorResolver.lambda$resolveSelectors$3(DiscoverySelectorResolver.java:62)
	at java.util.ArrayList.forEach(ArrayList.java:1257)
	at org.junit.jupiter.engine.discovery.DiscoverySelectorResolver.resolveSelectors(DiscoverySelectorResolver.java:61)
	at org.junit.jupiter.engine.JupiterTestEngine.resolveDiscoveryRequest(JupiterTestEngine.java:68)
	at org.junit.jupiter.engine.JupiterTestEngine.discover(JupiterTestEngine.java:61)
	at org.junit.platform.launcher.core.DefaultLauncher.discoverEngineRoot(DefaultLauncher.java:130)
	at org.junit.platform.launcher.core.DefaultLauncher.discoverRoot(DefaultLauncher.java:117)
	at org.junit.platform.launcher.core.DefaultLauncher.execute(DefaultLauncher.java:90)
	at com.microsoft.java.test.runner.junit5.CustomizedConsoleTestExecutor.lambda$executeTests$0(CustomizedConsoleTestExecutor.java:42)
	at com.microsoft.java.test.runner.junit5.CustomContextClassLoaderExecutor.invoke(CustomContextClassLoaderExecutor.java:29)
	at com.microsoft.java.test.runner.junit5.CustomizedConsoleTestExecutor.executeTests(CustomizedConsoleTestExecutor.java:38)
	at com.microsoft.java.test.runner.junit5.CustomizedConsoleLauncher.execute(CustomizedConsoleLauncher.java:29)
[...]

PR #539, opened 2018-12-21
Add parameter information for JUnit 5 test methods

### LINKED ISSUE #134: Could not run the tests through code lens if the method has parameters
1. Clone Junit projects from https://github.com/junit-team/junit5-samples
2. Open the junit5 project, e.g. junit5-maven-consumer
3. Trigger test from FirstTest class level: test passed
4. Trigger test from FirstTest method level: test skipped

image

LOG:
image

jdneo:
Fix #134 #526

### DIFF
--- a/java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestItemUtils.java
+++ b/java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestItemUtils.java
@@ -42,7 +42,7 @@ public class TestItemUtils {
         final Range range = parseTestItemRange(element);
         final String projectName = element.getJavaProject().getProject().getName();
 
-        return new TestItem(displayName, fullName, uri, range, level, kind, projectName, Collections.emptyList());
+        return new TestItem(displayName, fullName, uri, projectName, Collections.emptyList(), range, level, kind);
     }
 
     public static TestLevel getTestLevelForIType(IType type) {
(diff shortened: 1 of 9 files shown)


java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestItemUtils.java
```diff
@@ -15,6 +15,7 @@
 import com.microsoft.java.test.plugin.model.TestKind;
 import com.microsoft.java.test.plugin.model.TestLevel;
```
yaohaizh: TestItemUtils is mixed with JUnit5. could factor for different test platform.

jdneo: Good idea

jdneo: @yaohaizh Updated

Now, the code of parsing `IMethod` into `TestItem` is moved into each TestFramework searchers.

java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestItemUtils.java
```diff
@@ -26,70 +25,53 @@
...
+        final String displayName = element.getElementName();
+        final String fullName = parseTestItemFullName(element, level);
+        final String uri = JDTUtils.getFileURI(element.getResource());
+        final Range range = parseTestItemRange(element);
+        final String projectName = element.getJavaProject().getProject().getName();
 
-        return new TestItem(displayName, parseTestItemFullName(element, level),
-                JDTUtils.getFileURI(element.getResource()), parseTestItemRange(element, level), level, kind,
-                element.getJavaProject().getProject().getName());
+        return new TestItem(displayName, fullName, uri, range, level, kind, projectName, emptyList());
```
Eskibear: How about `Collections.emptyList()` instead of static importing?  Had a glance at the [ref](https://stackoverflow.com/questions/420791/what-is-a-good-use-case-for-static-import-of-methods), it looks like people are avoiding using static importing unless some constants are heavily used.

Issue #460, opened 2018-11-27
All tests that contain inner classes will be skipped if triggered from the test explorer


bsaby:
Result:
![image](https://user-images.githubusercontent.com/22470312/49061813-c321ab00-f24c-11e8-910c-449891fa9ea5.png)

Output
```
[ERROR] @@<TestRunner-{"name":"error", "attributes":{"message":"Failed to parse tests for suite com.example.project.JUnit4Test$A#testAssert1. Details: com.example.project.JUnit4Test$A#testAssert1.", "details":"java.lang.ClassNotFoundException: com.example.project.JUnit4Test$A#testAssert1\r\n\tat java.net.URLClassLoader.findClass(URLClassLoader.java:381)\r\n\tat java.lang.ClassLoader.loadClass(ClassLoader.java:424)\r\n\tat sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:349)\r\n\tat java.lang.ClassLoader.loadClass(ClassLoader.java:357)\r\n\tat java.lang.Class.forName0(Native Method)\r\n\tat java.lang.Class.forName(Class.java:264)\r\n\tat com.microsoft.java.test.runner.junit4.JUnit4RunnerUtil.getRequestForClasses(JUnit4RunnerUtil.java:80)\r\n\tat com.microsoft.java.test.runner.junit4.JUnit4RunnerUtil.createTestReferences(JUnit4RunnerUtil.java:45)\r\n\tat com.microsoft.java.test.runner.junit4.CustomizedJUnit4CoreRunner.run(CustomizedJUnit4CoreRunner.java:25)\r\n\tat com.microsoft.java.test.runner.junit4.JUnit4Launcher.execute(JUnit4Launcher.java:30)\r\n\tat com.microsoft.java.test.runner.Launcher.main(Launcher.java:53)\r\n"}}-TestRunner>.
[ERROR] 
.
[...]

PR #546, opened 2018-12-28
[Bug Fix] Cannot run the static inner test classes

### LINKED ISSUE #411: Will run all tests of outer class when trigger inner class in test explorer
Normal when triggering the test of inner class in a test file
But when trigger inner class in test explorer, it will run all tests including other methods under the outer class
image

jdneo:
Fix #411, #460


jdneo:
Should separate into two PRs

PR #556, opened 2019-01-02
Unescape the json string to get better format

### LINKED ISSUE #505: The test output is poorly readable
Build: 366

Result:
image

jdneo:
Resolve #505

### Before
before

### After
after

### DIFF
--- a/src/runners/baseRunner/BaseRunnerResultAnalyzer.ts
+++ b/src/runners/baseRunner/BaseRunnerResultAnalyzer.ts
@@ -14,18 +14,20 @@ export abstract class BaseRunnerResultAnalyzer {
     }
 
     public analyzeData(data: string): void {
-        logger.info(data);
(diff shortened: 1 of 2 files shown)


src/runners/baseRunner/BaseRunnerResultAnalyzer.ts
```diff
@@ -64,6 +66,14 @@ export abstract class BaseRunnerResultAnalyzer {
         };
     }
 
+    protected unescape(content: string): string {
+        return content.replace(/\\r/gm, '\r')
```
Eskibear: something like `s/(\\[rfntb])/&/g` would be better.

Eskibear: http://es5.github.io/x7.html#x7.8.4
table of SingleEscapeCharacter

Issue #524, opened 2018-12-14
Redesign the user experience about running/debugging with configuration


jdneo:
### Related Issues
#496, #476, #438 
#458, #372, #354, ...

### Description
Currently, the extension will generate the `launch.test.json` for users who want to run/debug with configuration for his test cases. While this approach has a lot of problems that hurt the user experices, including:
- No documents
- The schema is hard to understand
- Could not resolve variables such as `${workspaceFolders}`
- Ruin users' source control
- Obscure implementation logic

### Plan
Add several extension settings whose scope are workspace folder, the schema looks like below:
```js
"java.test.config": [
    {
        /* Specify the working directory when running the tests */
        "java.test.cwd": "${workspaceFolder}"

        /* Specify the command line arguments passed to the runner */
        "java.test.args": [ "" ]

        /* Specify the extra options and system properties for the JVM (e.g. -Xms<size> -Xmx<size> - D<name>=<value>) */
        "java.test.vmargs": [ "" ]

        /* Specify the environment variable when running the tests */
        "java.test.env": { "aaa": "bbb" }
    },
    {
        ...  // Another new config
    }
]
```

> Note: The first config in the array will always be picked as the default one.

jdneo:
Thank you for reporting @chuckb. Tracking issue: #602

PR #560, opened 2019-01-04
Remove test config related codes

jdneo:
First part of #524


jdneo:
Close it since we need to give users some time to migrate.

PR #566, opened 2019-01-08
Classify the tests according to their belonging workspace folder before loading the config

jdneo:
Part of #524


src/runners/runnerExecutor.ts
```diff
@@ -22,15 +20,14 @@ class RunnerExecutor {
...
-    private _preLaunchTask: cp.ChildProcess | undefined;
     private _runnerMap: Map<ITestRunner, ITestItem[]> | undefined;
 
     public initialize(javaHome: string, context: ExtensionContext): void {
         this._javaHome = javaHome;
         this._context = context;
     }
 
-    public async run(testItems: ITestItem[], isDebug: boolean = false, config?: IExecutionConfig): Promise<void> {
+    public async run(testItems: ITestItem[], isDebug: boolean = false, usingDefaultConfig: boolean): Promise<void> {
```
Eskibear: Why `isDebug` has default value while `usingDefaultConfig` doesn't have one?
Consider the case of the following calls:
run(a)
run(a, b)
run(a, b, c)

PR #567, opened 2019-01-08
Add a document for new setting

jdneo:
Document work for #524


runner-config.md
```diff
@@ -0,0 +1,44 @@
+To run the tests with a customized configuration, you can add a new setting with key named: `java.test.config`.
```
akaroml: Edit:
You can run tests with custom configuration. The configuration can be specified in `settings.json` using the key `java.test.config`.

runner-config.md
```diff
@@ -0,0 +1,44 @@
...
+
+> [How to create new settings in VS Code?](https://code.visualstudio.com/docs/getstarted/settings#_creating-user-and-workspace-settings)
```
akaroml: It would be great if we can provide extra instruction to guide users to put the configuration in the right place. In most cases, users may want to save the test config in `workspace settings`. So the instruction would be:

> It is recommended to save the test configuration as workspace settings. To open workspace settings, run command `Preference: Open Workspace Settings`. And click on the icon `{}` on the top right to show the source code of your settings.

runner-config.md
```diff
@@ -0,0 +1,44 @@
...
+        "env": { "key": "value" },
+    },
+    {
+        // Another configuration item...
+    },
+  ...
+]
+```
+
+> Note: The first configuration item in the array will be used as the default configuration when you trigger `Run Test` or `Debug Test`
```
akaroml: Having second thought regarding this. If a test configuration exists, then users have no way of running without a config. It's a risky move. I would suggest that we leave `Run Test` and `Debug Test` alone.

runner-config.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Property Details
+
+The value of `java.test.config` is an object array, each object is a configuration item, which may have following properties:
```
akaroml: It can be an object or an array of objects.

runner-config.md
```diff
@@ -0,0 +1,44 @@
...
+|---|---|---|
+| `name` | Specify the name of the configuration item | "" |
+| `workingDirectory` | Specify the working directory when running the tests | "${workspaceFolder}" |
+| `vmargs` | Specify the extra options and system properties for the JVM | [] |
+| `args` | Specify the command line arguments which will be passed to the test runner | [] |
+| `env` | Specify the extra environment variables when running the tests | {} |
+
+## Q & A
+**Q: How can I migrate to the new `java.test.config` setting?**
+A: You can create your own customized configuration according to the above document. Then remove or rename the `launch.test.json` in your workspace.
```
akaroml: customized configuration -> test configuration using the instruction above

Then the `launch.test.json` can be removed.

runner-config.md
```diff
@@ -0,0 +1,44 @@
...
+
+**Q: Can I keep using the original `launch.test.json` to run with customized configuration?**
+A: For now, yes. But the `launch.test.json` is deprecated and we will stop honor `launch.test.json` in the near future.
```
akaroml: Yea for now. But it's highly recommended that you work with the new schema. `launch.test.json` is deprecated and will be completely removed in the future.

runner-config.md
```diff
@@ -0,0 +1,44 @@
...
+
+**Q: What will happen if I both have `java.test.config` setting and `launch.test.json` in my workspace?**
+A: The extension will only take the configurations from `java.test.config` setting if it exists.
```
akaroml: `java.test.config` wins.

runner-config.md
```diff
@@ -0,0 +1,58 @@
...
+
+| Property Name | Description | Default Value |
+|---|---|---|
+| `name` | Specify the name of the configuration item | "" |
+| `workingDirectory` | Specify the working directory when running the tests | "${workspaceFolder}" |
+| `vmargs` | Specify the extra options and system properties for the JVM | [] |
+| `args` | Specify the command line arguments which will be passed to the test runner | [] |
+| `env` | Specify the extra environment variables when running the tests | {} |
+
+## Set the default configuration
```
akaroml: This part is a bit too verbose. I would recommend focusing on the following:
1. Introduce java.test.defaultConfig
2. Explain the special value `default`

There's no need to introduce how to "let the extension to set the default config" because it happens automatically.

runner-config.md
```diff
@@ -0,0 +1,63 @@
...
+
+If you do not want to select the configuration each time running the test cases. you can set the setting: `java.test.defaultConfig` to the name of the configuration you want to use.
```
akaroml: When you have custom configurations, you'll be asked to pick one everytime when running/debugging your test cases. To avoid that, you can use  `java.test.defaultConfig` and set its value to the name of a configuration.

If you want to bypass your custom configuration, you can use the built-in configuration named "default" which has the following definition:
```json
{
    "name": "default",
    "workingDirectory": "$(workspaceFolder)"
}
```

runner-config.md
```diff
@@ -0,0 +1,63 @@
...
+
+## The built-in configuration
```
akaroml: This section can be merged into the above one.

runner-config.md
```diff
@@ -0,0 +1,60 @@
...
+
+| Property Name | Description | Default Value |
+|---|---|---|
+| `name` | Specify the name of the configuration item | "" |
+| `workingDirectory` | Specify the working directory when running the tests | "${workspaceFolder}" |
+| `vmargs` | Specify the extra options and system properties for the JVM | [] |
+| `args` | Specify the command line arguments which will be passed to the test runner | [] |
+| `env` | Specify the extra environment variables when running the tests | {} |
+
+## Set the default configuration
```
akaroml: Suggest renaming this section to Custom Configurations

runner-config.md
```diff
@@ -0,0 +1,60 @@
...
+
+When you have custom configurations, you'll be asked to pick one everytime when running/debugging your test cases. To avoid that, you can use `java.test.defaultConfig` and set its value to the name of a configuration.
+
+If you want to bypass customizing the configurations, you can use the built-in configuration named "default" which has the following definition:
```
akaroml: Here what we want to bypass is the custom configurations but not the procedure of customizing... They have different meanings. I don't quite understand the meaning by looking at the sentence because customizing is not required by default.

The line can be changed to If you want to bypass any existing configurations, ...

PR #568, opened 2019-01-09
Maven extension is required for Test Runner

brunoborges:
If Maven for Java extension is not required, then this extension needs proper documentation on how to use it on a non-Maven project, such as a plain project created by the Java Dependency Viewer extension.

// @hexiaokai


jdneo:
@brunoborges Maven extension is not required. Currently, Test Runner only supports Maven/Gradle project.

The plain project is not supported yet. I'll update the document in another PR.

PR #569, opened 2019-01-09
Hint the user when the deprecated configuration is used

jdneo:
Part of work for #524

### DIFF
--- a/src/testConfigManager.ts
+++ b/src/testConfigManager.ts
@@ -2,8 +2,11 @@
 // Licensed under the MIT license.
 
 import * as fse from 'fs-extra';
+import * as opn from 'opn';
 import * as path from 'path';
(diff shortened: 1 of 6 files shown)


src/testConfigManager.ts
```diff
@@ -2,8 +2,11 @@
 // Licensed under the MIT license.
 
 import * as fse from 'fs-extra';
+import * as opn from 'opn';
```
Eskibear: use `vscode.open`

src/testConfigManager.ts
```diff
@@ -96,6 +99,30 @@ class TestConfigManager {
         }
         return selection.item;
     }
+
+    private async hintForDeprecatedUsage(): Promise<void> {
+        sendInfo('', { deprecatedConfigUsed: 1 });
```
Eskibear: properties should follow type [key: string]: string

PR #588, opened 2019-01-17
Fix the bug that status icon is not right in the inner test class

### LINKED ISSUE #586: A question status icon will be added as the third Code Lens automatically for inner class
### Build
416
### Steps:

1. Open a java file.
2. Trigger tests from the outer class codelens

OR
1. Open a java file and  and make sure you have the class with a status icon
2. Add a new inner class
3. Save file

### Result:
A question status icon will be added automatically for inner class
Screenshot1:
image

Screenshot2:
image

jdneo:
Fix #586

### DIFF
--- a/src/codeLensProvider.ts
+++ b/src/codeLensProvider.ts
@@ -77,9 +77,21 @@ class TestCodeLensProvider implements CodeLensProvider {
 
     private hasTestResult(test: ITestItem): boolean {
         if (test.level === TestLevel.Method) {
-            return testResultManager.hasResultWithFsPathAndFullName(Uri.parse(test.uri).fsPath, test.fullName);
+            return testResultManager.getResultDetails(Uri.parse(test.uri).fsPath, test.fullName) !== undefined;
         } else if (test.level === TestLevel.Class || test.level === TestLevel.NestedClass) {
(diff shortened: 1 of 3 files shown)


src/codeLensProvider.ts
```diff
@@ -76,12 +76,28 @@ class TestCodeLensProvider implements CodeLensProvider {
...
+                return testResultManager.getResultDetails(Uri.parse(test.uri).fsPath, test.fullName) !== undefined;
+            case TestLevel.Class:
+            case TestLevel.NestedClass:
+                const resultsInFsPath: Map<string, ITestResultDetails> | undefined = testResultManager.getResults(Uri.parse(test.uri).fsPath);
+                if (!resultsInFsPath) {
+                    return false;
+                }
+                const keyCollection: string[] = Array.from(resultsInFsPath.keys());
+                for (const child of test.children) {
+                    if (child.level !== TestLevel.Method) {
```
Eskibear: I assume the body of the `for` statement can be simplified as the following?
```ts
if (child.level === TestLevel.Method && keyCollection.indexOf(child.fullName) >= 0) {
    return true;
}
```

PR #601, opened 2019-01-28
Several small fixes

jdneo:
Several code changes including:
- Better tooltip content for navigating button in the test report
- Set the executor status at first in cleanup method
- Explicitly cast the `configSetting` to array when we know that it is.


jdneo:
Sure. Will separate the PR……

PR #615, opened 2019-02-15
Update runner-config.md

AlexAegis:


jdneo:
@AlexAegis Thank you for your contribution!

PR #627, opened 2019-02-24
Clear test result when test code changed #42

### LINKED ISSUE #42: Clear test result when test code changed
Currently, the test result stays when a user changes the actual test code. This could be misleading.

Please clear the test result whenever the test code is changed.

tonytamwk:
Fixed #42

### DIFF
--- a/src/testResultManager.ts
+++ b/src/testResultManager.ts
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
-import { Disposable, Uri, workspace, TextDocumentChangeEvent } from 'vscode';
+import { Disposable, TextDocumentChangeEvent, Uri, workspace } from 'vscode';
 import { ITestResult, ITestResultDetails } from './runners/models';
 
 class TestResultManager implements Disposable {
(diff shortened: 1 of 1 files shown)


jdneo:
@tonytamwk Thank you for your contribution.

From the changes, the new behavior will be: clear all the status icon in the Code Lens in the file, if it's modified.

This will fix #42, but I think we could do even better. For example, if the class has two methods: `a()` & `b()`, when `a()` is modified, only the status of `a()` get removed.

To achieve this, we might need to talk with the Java Language Server, send a request with the file uri and the modified range information to it, then let the Language Server to find the method which is being modified and return the name back to the TS client. 

This could be much harder than only changing the TS client part. Would you mind to have a try? Or you can let me finish the Java part, which is fine.

tonytamwk:
@jdneo Thank you for your comments.
Agree that detecting method change is a better approach and will have better usability. Let me try to implement this.

jdneo:
@tonytamwk The way to interact with the Language Server is not well documented so far, so please feel free to ask any question if you have. Meanwhile, I'm also pleased to help you finish the Java part, thanks.

jdneo:
Hi @tonytamwk,

Thank you for your contribution. I proposed a new way to mitigate the #42. More detailed information could be found here: https://github.com/microsoft/vscode-java-test/issues/42#issuecomment-503827198

Thank you again.

Issue #621, opened 2018-12-24
Spring Boot Test Fails with NoSuchBeanDefinitionException.


ecormaksin:
Any advice would be appreciated.

[provide a description of the issue]

##### Environment
- Operating System: macOS High Sierra 10.13.6
- JDK version: 1.8.0_151
- Visual Studio Code version: 1.30.1
- Java extension version: 0.5.0
- Java Debugger extension version: 0.16.0

##### Steps To Reproduce
1. Create Spring Boot project by following the page below. https://code.visualstudio.com/docs/java/java-spring-boot
2. Write test code with annotations below.
`@RunWith(SpringRunner.class)`
`@SpringBootTest`
`@ContextConfiguration(classes = BatchTestConfig.class)`
3. Run the test.

[attach a sample project reproducing the error]
[2018-12-24_TestReport.log](https://github.com/Microsoft/vscode-java-debug/files/2706758/2018-12-24_TestReport.log)

##### Current Result

NoSuchBeanDefinitionException occurs. But the test finishes with green on the Eclipse IDE.

##### Expected Result

The test finishes with green(passed).

jdneo:
Hi @ecormaksin,

Thank you for reporting the issue and give us the sample project. After some investigation, I think I might found out the root cause.

The root cause is that when the extension running the tests, it will read the `application.properties` under the target folder. But the content of it is the same as that in the `src` (not the one in `test`). After I copied the `application.properties` in the `test` to the `target` folder, the test cases can be run successfully.

This issue might be related to #534 

I'll think about how to make a fix and let you know.

Thank you.

jdneo:
More investigation result here: https://github.com/Microsoft/vscode-java-test/issues/534#issuecomment-466897041

jdneo:
Closing this issue and let's track it in #534

ecormaksin:
I apoligize for too late reply. Thank you so much!

Issue #534, opened 2018-12-19
Test context reading 'wrong' application properties file


pablo-parra:
In a Spring Boot sample project using a MySQL db as main data source and H2 as test data source, when launching the test using Java Test Runner extension the application properties file read by the app is the one located in the "main" package instead the one located in the "test" package. Due to that all the test fail as are not finding the expected infrastructure.

The test of the project are working fine in Eclipse.

![application_properties_issue](https://user-images.githubusercontent.com/18135212/50227243-9bce8180-03a5-11e9-8eaa-d7144a1c3e30.png)

rutcreate:
Same problem here. If there is any workaround please share.

For my case, I use DBRider to test database. It will delete tables in database before or after start test case. From this problem cause dev's database tables will be deleted instead of test's database.

jdneo:
@pablo-parra @rutcreate 

It might have a better solution to this issue. For now, you can try the following workaround:

1. Check if the `application.properties` is in `/target/test-classes/`
2. If yes, then open your `launch.test.json` file, in `env` field, add the following setting: `"spring.config.location": "classpath:/target/test-classes/"`
3. Do not forget to set the `default` field as well, otherwise the extension will not using the configuration as the default.

Here is an example:

```json
{
    "run": {
        "default": "todo-app-java-on-azure",
        "items": [
            {
                "name": "todo-app-java-on-azure",
                "projectName": "todo-app-java-on-azure",
                "workingDirectory": "c:\\Users\\sheche\\Desktop\\todo-app-java-on-azure-master",
                "args": [
                ],
                "vmargs": [],
                "env": {
                    "spring.config.location": "classpath:/target/test-classes/"
                },
                "preLaunchTask": ""
            }
        ]
    },
    // ...
}
```

Please let me know if this could help with your problem.

Thanks

pablo-parra:
Hi @jdneo, I have followed the steps you mention and it's working now 👍. 
I guess it is something that could be done automatically by the extension or at least should be mentioned in the "Usage" section of the extension landing page.

Thanks for your help!! :)

yaohaizh:
@jdneo, please document this either on readme or wiki page.

pablo-parra:
[...]

PR #629, opened 2019-02-26
[bugfix] Put the target/test-class before target/class when calculating the classpath

jdneo:
To address #534 & #621 

The root cause is that the `application.properties` in `src` is loaded before that in `test`

The `target/test-class` must come before `target/class`.


Spitfire1900:
I think this is still occurring, how does one report the classpath when running a Test?

PR #668, opened 2019-03-29
Add no-response bot

jdneo:


.github/no-response.yml
```diff
@@ -0,0 +1,4 @@
+daysUntilClose: 14
+responseRequiredLabel: need more info
```
Eskibear: "needs" or "need"? I prefer to align with VSCode repo to use "needs"

Eskibear: and the label is about **the issue**, the issue **needs** more info.

PR #683, opened 2019-05-01
Supoprt for jqwiq Property based test annotation.

Polish-Civil:


jdneo:
Hi @Polish-Civil,

Thank you for your contribution. 

For the PR itself, I'm afraid this might not be a good way to add different annotation support since there will be a lot of third-party annotation and it's impossible to hard-code them all.

But One thing I noticed is that the `@Property` [is annotated by another Junit 5 official annotation: `org.junit.platform.commons.annotation.testable`](https://github.com/jlink/jqwik/blob/master/api/src/main/java/net/jqwik/api/Property.java#L30) and this could be a good start point to solve the problem. Would you mind to investigate more and update your PR?

owickstrom:
@jdneo I was just going to add an issue for that, adding `org.junit.platform.commons.annotation.Testable` to the list of accepted annotations, but you beat me to it here. It does indeed seem to remove the need for third-party additions, and even the existing `Test` and `TestFactory`. Looking at its Javadoc:

> `@Testable` is used to signal to IDEs and tooling vendors that the annotated or meta-annotated element is testable.
>
> In this context, the term "testable" means that the annotated method or class can be executed by a TestEngine as a test or test container on the JUnit Platform.
>
> [...]
> 
> `@Testable` will typically be used as a meta-annotation in order to create a custom composed annotation that inherits the semantics of `@Testable`. For example, the `@Test` and `@TestFactory` annotations in JUnit Jupiter are meta-annotated with `@Testable`.

jdneo:
Hey @owickstrom,

Thank you for the feedback. Yes just as you have said. Supporting the testable annotation should be better than support `TestTemplate`, `TestFactory`, etc...

I'll take a look at it and try to support it ASAP

Thanks

jdneo:
Closing this PR since the testable feature has been supported in https://github.com/microsoft/vscode-java-test/pull/980. (Though we may still have some problem triggering the test from explorer, but it should be fine when using the Code Lens).

@Polish-Civil still want to say thank you for your contribution!

PR #711, opened 2019-05-23
fix: print the unexpected exceptions to the output channel

### LINKED ISSUE #537: TestNG test cases will be skipped
According to https://github.com/Microsoft/vscode-java-test/issues/504#issuecomment-445131893 by @bsaby 

The TestNG test cases will be skipped if triggering from workspace root level and workspace folder level.

## Root cause
In `com.mkyong.testng.examples.group.TestGroup`, some of the codes are commented. Release them will make the runner running successfully.

## How to fix
I will investigating if there is a way to let the runner run other tests no matter the usage of `dependsOnGroups`

jdneo:
Fix #537 

### Previous Behavior
The extension won't show anything when exception happens

### Now
Give the user a hint
demo

### DIFF
--- a/src/runners/runnerExecutor.ts
+++ b/src/runners/runnerExecutor.ts
@@ -17,6 +17,8 @@ import { JUnit4Runner } from './junit4Runner/Junit4Runner';
 import { JUnit5Runner } from './junit5Runner/JUnit5Runner';
 import { ITestResult } from './models';
 import { TestNGRunner } from './testngRunner/TestNGRunner';
+import { OPEN_OUTPUT_CHANNEL } from '../constants/dialogOptions';
+import { showOutputChannel } from '../commands/logCommands';
 
 class RunnerExecutor {
     private _javaHome: string;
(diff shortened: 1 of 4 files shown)


src/runners/runnerExecutor.ts
```diff
@@ -75,13 +77,17 @@ class RunnerExecutor {
...
             }
-        } catch (error) {
-            window.showErrorMessage(`${error}`);
-            testStatusBarProvider.showFailure();
-        } finally {
             testStatusBarProvider.showTestResult(finalResults);
             testCodeLensProvider.refresh();
             testReportProvider.update(finalResults);
+        } catch (error) {
+            window.showErrorMessage(`${error}`, OPEN_OUTPUT_CHANNEL).then((choice: string | undefined) => {
```
Eskibear: `${error}` = error ?

jdneo: Good catch

PR #756, opened 2019-07-24
feat: Support Junit 5 meta-annotations and composed annotations

### LINKED ISSUE #737: Support Junit 5 Meta-Annotations and Composed Annotations
Reference: https://junit.org/junit5/docs/current/user-guide/#writing-tests-meta-annotations

jdneo:
resolve #737 

demo

By searching the annotation hierarchy, we can resolve the meta-annotation scenario.

### DIFF
--- a/java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/searcher/TestNGTestSearcher.java
+++ b/java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/searcher/TestNGTestSearcher.java
@@ -43,7 +43,7 @@ public class TestNGTestSearcher extends BaseFrameworkSearcher {
                 return false;
             }
             for (final String annotation : this.testMethodAnnotations) {
-                if (TestFrameworkUtils.hasAnnotation(method, annotation)) {
+                if (TestFrameworkUtils.hasAnnotation(method, annotation, false /*checkHierarchy*/)) {
                     return true;
                 }
             }
--- a/java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestFrameworkUtils.java
+++ b/java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestFrameworkUtils.java
@@ -19,20 +19,20 @@ import com.microsoft.java.test.plugin.searcher.TestNGTestSearcher;
 
 import org.eclipse.jdt.core.IAnnotatable;
 import org.eclipse.jdt.core.IAnnotation;
-import org.eclipse.jdt.core.ICompilationUnit;
+import org.eclipse.jdt.core.IJavaProject;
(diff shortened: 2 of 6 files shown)


java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestFrameworkUtils.java
```diff
@@ -52,51 +52,93 @@ public static TestItem resolveTestItemForClass(IType type) throws JavaModelExcep
...
-    public static boolean hasAnnotation(IMember member, String annotation) {
-        return getAnnotation(member, annotation).isPresent();
+    /**
+     * Check the given {@link IMember} has annotated with the given annotation string.
+     *
+     * @param member the {@link IMember} to search.
+     * @param annotationToSearch The annotation string.
+     * @param checkHierarchy Specify whether to search the whole annotation hierarchy.
+     */
+    public static boolean hasAnnotation(IMember member, String annotationToSearch, boolean checkHierarchy) {
```
Eskibear: Would `Inherited` be more proper than hierarchy? e.g. `checkInherited`

Eskibear: `includeInherited` looks even better.

java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/searcher/TestNGTestSearcher.java
```diff
@@ -43,7 +43,7 @@ public boolean isTestMethod(IMethod method) {
                 return false;
             }
             for (final String annotation : this.testMethodAnnotations) {
-                if (TestFrameworkUtils.hasAnnotation(method, annotation)) {
+                if (TestFrameworkUtils.hasAnnotation(method, annotation, false /*checkHierarchy*/)) {
```
Eskibear: Actually you can simply overload `hasAnnotation`, to avoid changing these statements. 
```
boolean hasAnnotation(method, anno) {
  return hasAnnotation(method, anno, false);
}

boolean hasAnnotation(method, anno, boolean) {
 // ...
}
```

java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestFrameworkUtils.java
jdneo: Hmm, actually the annotation we treated here is not inherited. That's why it is called meta-annotation.

Eskibear: exactly, not inherited. From the saying meta-annotation, it's annotation of an annotation, and it is, more like aggregation/combination.  
Anyway, use whatever proper name you like.

PR #763, opened 2019-08-02
feat: Support JUnit 5 TestTemplate

jdneo:
demo


java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/searcher/JUnit5TestSearcher.java
```diff
@@ -41,7 +41,7 @@ public JUnit5TestSearcher() {
         super();
         this.testMethodAnnotations = new String[] { "org.junit.jupiter.api.Test",
             "org.junit.jupiter.params.ParameterizedTest", "org.junit.jupiter.api.RepeatedTest",
-            "org.junit.jupiter.api.TestFactory" };
+            "org.junit.jupiter.api.TestFactory", "org.junit.jupiter.api.TestTemplate" };
```
testforstephen: can this be a user setting?

jdneo: Hmm, some reasons that I don't want to expose it to a user setting.
- Different annotation has different restrictions for the method signature, for example, some requires that the method must return void but some are not.
- User may misconfigure an annotation which is not testable actually.
- Refer to the IntelliJ & Eclipse implementation, it's hard coded as well.

PR #783, opened 2019-08-21
chore: Migrate vscode to @types/vscode and vscode-test

jdneo:


package.json
```diff
@@ -317,6 +316,7 @@
         "@types/mocha": "^2.2.48",
         "@types/node": "^6.0.110",
         "@types/pug": "^2.0.4",
+        "@types/vscode": "^1.32.0",
```
Eskibear: Suggest to fix the version by removing ^. Otherwise you might install the latest version and use the new APIs without knowing it.
It should match the engine.

jdneo: @Eskibear Addressed

PR #791, opened 2019-09-04
chore: Update the build pipelines

jdneo:
- Update launch.json
 - Add apache.common.lang3
- Update target platform

### DIFF
--- a/java-extension/com.microsoft.java.test.plugin/pom.xml
+++ b/java-extension/com.microsoft.java.test.plugin/pom.xml
@@ -37,40 +37,6 @@
                 </executions>
             </plugin>
             <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-dependency-plugin</artifactId>
-                <version>3.1.1</version>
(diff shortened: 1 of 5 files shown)


java-extension/com.microsoft.java.test.plugin/target.target
```diff
@@ -24,4 +29,4 @@
             <repository location="http://download.eclipse.org/tools/orbit/R-builds/R20170516192513/repository"/>
```
Eskibear: add `<unit id="org.apache.commons.lang3" version="3.1.0.v201403281430"/>` here to acquire

java-extension/com.microsoft.java.test.plugin/pom.xml
```diff
@@ -62,6 +62,11 @@
                             <artifactId>gson</artifactId>
                             <version>2.7</version>
                         </artifactItem>
+                        <artifactItem>
```
Eskibear: no need to use m2 dependency, see below

PR #801, opened 2019-09-10
refactor: Allow the test runner to accept the debug launch configuration

jdneo:


src/runners/runnerScheduler.ts
```diff
@@ -26,14 +27,14 @@ class RunnerScheduler {
...
-    private _runnerMap: Map<ITestRunner, ITestItem[]> | undefined;
+    private _runnerMap: Map<BaseRunner, ITestItem[]> | undefined;
 
     public initialize(javaHome: string, context: ExtensionContext): void {
         this._javaHome = javaHome;
         this._context = context;
     }
 
-    public async run(testItems: ITestItem[], isDebug: boolean, node?: TestTreeNode): Promise<void> {
+    public async run(testItems: ITestItem[], isDebug: boolean, node?: TestTreeNode, launchConfiguration?: DebugConfiguration): Promise<void> {
```
Eskibear: Consider to group the args into one “options” object? Benefit is you won’t need to change the signature any more

testforstephen: TestTreeNode is data model at ui layer, it's better not to use it directly in the business logic layer.

PR #802, opened 2019-09-11
chore: Download junit4-runtime and add it into the bundle list

jdneo:


gulpfile.js
```diff
@@ -69,3 +73,33 @@ function isWin() {
...
+                return extensionString.substring(0, extensionString.lastIndexOf('/') + 1) + fileName;
+            }
+            return extensionString;
+        });
+
+        fs.writeFileSync('./package.json', JSON.stringify(packageJsonData, null, 4));
+    }
+}
+
+function findNewPDEPlugin(fileName) {
```
Eskibear: should better add short comments for it, about the implication that PDE plugins are named as `<name>_<version>.jar`

PR #806, opened 2019-09-12
fix: Show error messages when error happens

jdneo:


jdneo:
Thanks for reviewing. Have a good holiday!

PR #834, opened 2019-10-12
fix: Replace the port number instead of appending a new one

### LINKED ISSUE #829: Duplicated port parameter when launch Junit tests

jdneo:
fix #829


Vigilans:
What if "-port" parameter does not exist?

Seems that if server returns no "-port" parameter, client will also not add one for it.

PR #839, opened 2019-10-18
refactor: Better handling the debug session

jdneo:


src/runners/baseRunner/BaseRunner.ts
```diff
@@ -42,52 +38,63 @@ export abstract class BaseRunner implements ITestRunner {
...
+                            if (data.length > 0) {
+                                this.testResultAnalyzer.analyzeData(data);
+                            }
+                            return resolve(this.testResultAnalyzer.feedBack());
+                        }
+                    }),
+                );
+            });
+        }, ((reason: any): any => {
+            logger.error(`${reason}`);
```
Eskibear: reason

jdneo: Seems I need to refine the logger's APIs... Will do it in the following PRs

Issue #126, opened 2018-02-12
show test status on test explorer nodes


ansyral:
Currently we don't show test status on nodes of test explorer, so when user choose to run all tests from test explorer, if some tests failed, they can open test report to check the failed tests but it is hard for them to locate the file that contains the failed tests.

thoni56:
This could be done by providing an adapter for the [Test Explorer UI](https://marketplace.visualstudio.com/items?itemName=hbenl.vscode-test-explorer).

PR #844, opened 2019-10-21
Easy "Re-Run Test" capability

JasonHenriksen:
This PR adds a status bar at the bottom of the page that will re-run the last run test with a single button push. 

I often use a unit test as something I re-run many times to exercise or explore a unit of code.  In that case the pass/fail report is not of much use.   However, it is an impedance to have to switch back and forth between the debugger and the test explorer to re-run each test.  With this PR, I never have to leave the debug page, I can simply kick off the test again and again.

This is also useful in situations were I have kicking off a repeated task such as a load test or when I'm setting break points attempting to debug code that I'm not actually in the process of changing.

I'm happy to do whatever updates are necessary to help this PR be accepted.

I'd also like to point out that I had a problem in the runnerScheduler on line 158 that I had to simply comment out.  Whatever I can do to repair that issue, I'm happy to help with.

Thank you for a great extension!  Your TestNG support allowed me to move from IntelliJ to full time VS Code usage!


jdneo:
Hey @JasonHenriksen, thank you for your contribution, could you please sign the CLA first? The guidance can be just found in the comment.

Meanwhile, could you please change the format to use 4 spaces as the indention?

JasonHenriksen:
OK.  I think that did it.  I'll be around for another half hour if there's anything else quick I can get lined up.

src/runners/runnerScheduler.ts
```diff
@@ -153,8 +155,16 @@ class RunnerScheduler {
             case TestKind.JUnit:
             case TestKind.JUnit5:
                 return new JUnitRunner(this._context.extensionPath);
-            case TestKind.TestNG:
-                return new TestNGRunner(this._context.extensionPath);
+
+            // As of 22:40 on Oct 20 this line is not compiling on HEAD
```
jdneo: Will reload the VS Code move this error away?

JasonHenriksen:
mind=blown
restart vscode did resolve that error.  putting that back in place and trying the CI again.

jdneo:
The CI is reporting just some linting errors. You can leverage the [TSLint](https://marketplace.visualstudio.com/items?itemName=ms-vscode.vscode-typescript-tslint-plugin) extension to help fix them

jdneo:
BTW, just curious the decision to put the rerun button into the status bar. It's working but might have some potential problem. Especially when other extensions are sending their message to their status bar component. (You will observe the position of the rerun button is keeping changing)

Do you think put it to the test explorer is a good idea? Or even we store the launch configuration into the launch.json for each execution. Then the user can only hit `F5` to rerun the last execution.

JasonHenriksen:
I think I've fixed the linting errors, but the java error I'm not sure what to do about.
Any assistance is appreciated, though I'll probably have to look at it tomorrow.  ( it's midnight here )

jdneo:
That's fine. No need to be hurry. Have a good rest.

JasonHenriksen:
thanks.  Please let me know if you can see what's wrong with that error and I'll take a stab at it tomorrow.

JasonHenriksen:
I looked at this again this morning.  I feel like this the regression test is failing due to some environment problem.  I didn't change anything to do with the java access configuration.   If you guys have any advice I'm happy to follow it, but I don't think I can do anything more to get this to work without some guidance.

JasonHenriksen:
> BTW, just curious the decision to put the rerun button into the status bar. 
> It's working but might have some potential problem. Especially when other 
> extensions are sending their message to their status bar component. 
> (You will observe the position of the rerun button is keeping changing)

The changing position is not a concern because it says. "Re-Run".  It's honestly nice because it's easy to find and click.  

> Do you think put it to the test explorer is a good idea? 

No, the whole problem being avoided is that you have to constantly click back and forth between windows when you just want to re-run the test you just completed.

> Or even we store the launch configuration into the launch.json for each execution. 
> Then the user can only hit `F5` to rerun the last execution.

There does not seem to be any launch configuration for tests.  
Pressing F5 attempts to launch the application, not run the test.  It doesn't seem like a good idea to put tests into launch configurations anyway since tests are different.  Sometimes I'll need to run the first test a handful of times, then move on to the next test and so on.   Building launch configurations would leave behind a trail of unused configuration settings after the work was done.

I'd be open to adding a "Re-Run Last Test" key binding.  ( Maybe Shift-F5 or something? )  But I'm going to wait until I can get past this TravisCI problem so I'm sure I know that I can get the fix in when it's done.   ( The lack of this button is a real hassle for me.  I'm testing a codegenerate and I'm re-running this test about twice a minute all day long. )

jdneo:
Hi @JasonHenriksen,

Don't worry about the CI failure, it's some linting errors:

```
ERROR: /home/travis/build/microsoft/vscode-java-test/src/extension.ts:26:1 - Import sources within a group must be alphabetized.
ERROR: /home/travis/build/microsoft/vscode-java-test/src/commands/explorerCommands.ts:12:1 - Import sources within a group must be alphabetized.
ERROR: /home/travis/build/microsoft/vscode-java-test/src/commands/explorerCommands.ts:14:1 - Consecutive blank lines are forbidden
ERROR: /home/travis/build/microsoft/vscode-java-test/src/commands/explorerCommands.ts:51:32 - " should be '
ERROR: /home/travis/build/microsoft/vscode-java-test/src/commands/explorerCommands.ts:52:32 - " should be '
ERROR: /home/travis/build/microsoft/vscode-java-test/src/commands/runFromCodeLens.ts:7:1 - Import sources within a group must be alphabetized.
ERROR: /home/travis/build/microsoft/vscode-java-test/src/commands/runFromCodeLens.ts:9:1 - Consecutive blank lines are forbidden
ERROR: /home/travis/build/microsoft/vscode-java-test/src/commands/runFromCodeLens.ts:19:32 - " should be '
ERROR: /home/travis/build/microsoft/vscode-java-test/src/commands/runFromCodeLens.ts:20:32 - " should be '
ERROR: /home/travis/build/microsoft/vscode-java-test/src/commands/runFromReTryStatusBar.ts:7:1 - Import sources within a group must be alphabetized.
ERROR: /home/travis/build/microsoft/vscode-java-test/src/commands/runFromReTryStatusBar.ts:9:1 - Consecutive blank lines are forbidden
ERROR: /home/travis/build/microsoft/vscode-java-test/src/commands/runFromReTryStatusBar.ts:11:43 - Type assertion using the '<>' syntax is forbidden. Use the 'as' syntax instead.
ERROR: /home/travis/build/microsoft/vscode-java-test/src/commands/runFromReTryStatusBar.ts:11:59 - missing whitespace
ERROR: /home/travis/build/microsoft/vscode-java-test/src/commands/runFromReTryStatusBar.ts:11:83 - " should be '
ERROR: /home/travis/build/microsoft/vscode-java-test/src/commands/runFromReTryStatusBar.ts:12:35 - Type assertion using the '<>' syntax is forbidden. Use the 'as' syntax instead.
ERROR: /home/travis/build/microsoft/vscode-java-test/src/commands/runFromReTryStatusBar.ts:12:46 - missing whitespace
ERROR: /home/travis/build/microsoft/vscode-java-test/src/commands/runFromReTryStatusBar.ts:12:70 - " should be '
```

To fix them, you can install the [TSLint](https://marketplace.visualstudio.com/items?itemName=ms-vscode.vscode-typescript-tslint-plugin) extension, then open the related files shows above in VS Code. Then you can leverage the `quickfix` provided by the `TSLint` extension to fix them. 

But before that, I would like to discuss about the whole UX first.

Adding a keybinding is a good idea. actually you can register this re-run command into the `package.json`, then users can find them in the keyboard shortcut page:

image

In that page the user can even configure the shortcut according their own preference. With the shortcut supported, I think your problem will be addressed no matter we have the re-rurn status bar or not, right?

I'm not in favor of putting the button into the status bar because so far I cannot find a good example(other testing related extension) doing that. And you can see that the implementation is somehow duplicated with the result status bar that we already have. 

Feel free to let me know your thoughts about this.

Thanks.

jdneo:
Hi @JasonHenriksen,

Do you need any help for this PR?

JasonHenriksen:
Sorry for the delay, I can only really work on this at night when my kids are in bed or I'm not otherwise engaged.  

I can handle the TS Linting and I agree about the keyboard shortcut being important.

But I still prefer the status bar button for discover-ability.    Once we reach an agreement on UI I'll put in a fix to wrap this up.

As for an example, this plugin already has a status bar that gives the pass fail count.  I would have preferred to have another icon in that bar that would re-run the test, but the whole bar only seems able to launch one action when clicked.  That's why I added a second status item.  Personally, a re-run button is way more helpful than a "open the report" button.

How would you feel about a setting toggle that shows or hides the re-run status bar.  That way, if people don't like it, they can turn it off.   We could even have a toggle for both status' since I'd be happy not to waste the status space on a pass/fail count when I could just goto the report from the explorer.   What I really need is any easy-to-discover way to re-run the test.  The keybinding is definitely needed, but I would never have thought to go looking in the keybinding settings to find it. 
( Though maybe a label on the explorer that says "Press Ctrl-F5 to re-run last test" would suffice? )

How would you prefer to proceed with it:
- Settings to toggle status visibility
- Label in explorer for keybinding
- Some other idea?

Thanks for the discussion, 
- Jason

jdneo:
Thank you for these great ideas! So far I'm okay with that!

I'll discuss the UX with my teammates to see if they have any input and let you know the results.

Thank you for the efforts!

jdneo:
Hi @JasonHenriksen,

Could we first add the `re-run` feature as a command, and let the users to configure the keyboard shortcut by their own preference?

Adding the `re-run` button into `Test Explorer` and `Test Result Page` is safe since these two components are fully controlled by the Test Runner itself. While the status bar is the place where all extensions share. More importantly, the `re-run` button is a command not a status. That's why I feel we should be cautious to put it there.

But this does **not** mean that we won't add this short-cut to the status bar anymore. I'd like to first provide this feature with a small accessible places. During the users are using and finding this features, we will then receive the feedback from them. If we find lots of them want to put it into the status bar, then it will be a good chance to go a step further.

At least with the keyboard shortcut binding, your problem can be somehow unblocked. Does that sound good to you?

JasonHenriksen:
Thanks for your feedback.  I am enjoying the discussion of how best to approach this.

So I'll do as you ask, but I think that status bars as action is pretty common:
- test launches a report.
- the git extension has 2 status bars.  One starts a branch process, the other does a code sync
- the happy face opens a window to launch a tweet

That said, I agree with wanting to avoid clutter.  So I think I'll make the Alt-Text of the status bar that currently says "View Report" read "Click to View Report - Alt-F5 to re-run last test"  That way people have some hope of discovering the capability.   I'll also add a re-run button to the explorer.

( I'm also not clear why the report is separate from the explorer.  It would be way easier to not have the report at all and just add the Passed/Failed icon beside the test in the explorer.  But that's a different effort entirely )

Post again soon,

- Jason

Thanks again for the time and discussion.   I'll update the PR sometime in the next few days.

jdneo:
Hey @JasonHenriksen,

> ( I'm also not clear why the report is separate from the explorer. It would be way easier to not have the report at all and just add the Passed/Failed icon beside the test in the explorer. But that's a different effort entirely )

Yes, put the status icon aside the node in Test Explorer is one work item that we will take a look into. So far, we have an issue to track that: #126

> So I think I'll make the Alt-Text of the status bar that currently says "View Report" read "Click to View Report - Alt-F5 to re-run last test"

Yes, feel free to make this change. Meanwhile, after the PR is merged, I'll create an issue to ask users where they hope the re-run button they can access from. That will also be a way to expose the feature and collect the feedback.

JasonHenriksen:
I tweaked it and got the lint to pass.  Alt-T will re-run the test and I commented out the status bar.
I also added a button on the explorer but I reused one of the existing icons.
I hope it helps out!

jdneo:
Hi @JasonHenriksen,

Thank you for update the PR. 

Don't worry about the icon and translations, I'll work with them.

I've to say sorry that during this period, I update several commits to change the code base. So far this PR just revert those changes back. So would you mind to correct them before I can take a look?

JasonHenriksen:
I think I had done the merge as requested, but this is still hanging out. 
Were there any other problems with it or other work that needed to be done?

jdneo:
Hi @JasonHenriksen

I guess you resolve the conflicts by mistake when merging the branch.

For example, in `package.json`. The setting `java.test.saveAllBeforeLaunchTest` is actually removed in master branch, but it come back again in this PR.

jdneo:
Hi @JasonHenriksen,

My apologize that we could not work together to get the PR merged. It's appreciated for your work and I would like to say thank you for your contribution.

I have raised another PR to add this capability to Test Runner in #1031. The re-run command will be available in the next release.

At last, thank you again for your work and the idea.

Thanks,
Sheng

PR #846, opened 2019-10-22
Migrate the icons to vscode-icon in Test Explorer

jdneo:
### Dark
image

### Light
image

### Dark
image

### Light
image


jdneo:
Related issue here: https://github.com/microsoft/vscode/issues/31466.

We can update the reference once the upstream issue is resolved.

PR #847, opened 2019-10-22
Migrate the Code Lens status icon to Octicons

### LINKED ISSUE #845: Use octicon to replace emojis
This is to request replacing the emojis in the editor view:
image

VS Code now supports oct icons, and by adopting them, the test status looks more consistent and appealing in the editor window.
https://code.visualstudio.com/api/references/icons-in-labels
https://octicons.github.com/

jdneo:
Fix #845.

Appearance:
image

image

### DIFF
--- a/src/codelens/TestCodeLensProvider.ts
+++ b/src/codelens/TestCodeLensProvider.ts
@@ -8,7 +8,6 @@ import { ITestItem, TestLevel } from '../protocols';
 import { ITestResult, ITestResultDetails, TestStatus } from '../runners/models';
 import { testResultManager } from '../testResultManager';
 import { searchTestCodeLens } from '../utils/commandUtils';
-import { isDarwin } from '../utils/platformUtils';
 
 export class TestCodeLensProvider implements CodeLensProvider {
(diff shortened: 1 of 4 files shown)


akaroml:
Nice. How about the cases that show a question mark?

src/codelens/TestCodeLensProvider.ts
```diff
@@ -139,7 +138,7 @@ export class TestCodeLensProvider implements CodeLensProvider {
             return '?';
```
akaroml: Does vscode has an icon to represent the unknown state?

jdneo: A related one might be:

image

But looks too complicated.

Or maybe we can just show nothing if the state is unknown.

akaroml: $(question)

jdneo: Per offline discussion. We will leave the question icon as what it is.

jdneo:
Result in VS Code 1.40.0 stable

image

Issue #835, opened 2019-10-13
Often get "A test session is currently running. Please wait until it finishes."


ZhengguanLi:
All I have to do is start my VSCode often...

jdneo:
Hey @ZhengguanLi,

I'll take a look to this issue. For now, you can use the command: `Java: Cancel Test Job` as a workaround.

calvin1864:
This happens reliably for me when I run tests when the project has a compiler error.

Clicking 'Abort' on the 'Build Failed' seems to leave the test job running without hope for completion.

This may be a different use case than OP's since it doesn't happen on startup, only when I trigger tests

jdneo:
Hi @all.

Thank you for the feedback. I've noticed this issue and try to fix it in: #839. The fix should be released in next week.

jdneo:
Thank you @calvin1864 for the information.

I'll close this issue for now. If the problem still occurs, please let me know.

Thanks.

shiroukamui:
> Hey @ZhengguanLi,
> 
> I'll take a look to this issue. For now, you can use the command: `Java: Cancel Test Job` as a workaround.as a workaround.as a workaround.

Hi @jdneo, 
Excuse me but what exactly do you mean with the command `Java: Cancel Test Job`? this isn't clear enough since it is not a valid command.
Thanks.

jdneo:
@shiroukamui  What do you mean by: 

> it is not a valid command.

PR #849, opened 2019-10-23
Issue #835:  Making sure vscode cleans up the running test 

calvin1864:
If a build fails and the user selects to Abort the running tests, the cleanup method is never called and the isRunning boolean is stranded as true.

This forces cleanup to occur if the user abandons their test run to fix build issues

### DIFF
--- a/src/runners/runnerScheduler.ts
+++ b/src/runners/runnerScheduler.ts
@@ -51,6 +51,7 @@ class RunnerScheduler {
                     'Proceed',
                     'Abort');
                 if (ans !== 'Proceed') {
+                    this.cleanUp(false);
                     return;
                 }
             }


jdneo:
Hi @calvin1864,

Thank you for providing this PR. This remind me that I need to do more refactoring work about the Scheduler.

The background is that previously, the Test Runner will launch a Java process by itself to run the tests. While recently I'm doing the refactoring work to migrate the launch task to Java Debugger. What the Test Runner does is just pass a launch configuration to the Java Debugger.

In #850, I removed the setting `java.test.forceBuildBeforeLaunchTest` and its related handling logic, because the users can use the same setting in debugger instead(`java.debug.settings.forceBuildBeforeLaunch`).

So the potential bug we mentioned here should disappear because the codes are actually removed.

But anyway, I still want to thank you for finding this out! I'll keep an eye on this issue to see if there are other cases to cause the bug.

_BTW, if you are interested, Please feel free to have a try with the latest build. You can pull the latest bits from the master branch._

PR #875, opened 2019-11-11
fix: Append the port args when it's not set

jdneo:
fix: https://github.com/testforstephen/vscode-pde/issues/12


src/runners/baseRunner/BaseRunner.ts
```diff
@@ -61,6 +61,10 @@ export abstract class BaseRunner implements ITestRunner {
             });
         });
 
+        // Run from integrated terminal will terminate the debug session immediately after launching,
+        // So we force to use internal console here to make sure the seesion is still under debugger's control.
```
testforstephen: typos. **seesion**

PR #885, opened 2019-11-27
Remove the encoding guess logic

jdneo:
Instead of guessing the encoding. Users should provide the encoding settings by their own through the `vmargs` in `java.test.config`


jdneo:
Job fail due to Font Awesome download link broken: https://github.com/FortAwesome/Font-Awesome/issues/15845. I'll merge this PR for now.

PR #898, opened 2019-12-04
feat: Hint for unsupported test usage

jdneo:


jdneo:
Close for more investigation needed.

PR #942, opened 2020-02-21
fix: Improve the test report style

### LINKED ISSUE #933: Wired looking after resizing the column
* text overlap: passed and 118ms
* text overflow: name of classes in table header

Screen Shot 2020-02-04 at 10 54 43 AM

### LINKED ISSUE #938: Horizontal scrollbar too low
Please see the screenshot below.  To scroll right/left I first have to scroll down -> then right -> then up and repeat 🔁.  It's very tedious.

image

### LINKED ISSUE #939: Test report UI Feedback: Overlapping Text, Unformatted Values
1. The timings are **on top** of the passing label when the values are large enough
1. Can we **format** the **ms** results? i.e. 3,038ms vs 3038ms or maybe even 3.04s instead of 3083ms
1. Can't read full test name and hovering of it doesn't display it.  Can we **wordwarp** the test name? Perhaps that will be a problem as there are no spaces but there's probably an answer for that somewhere on google/stackoverflow.

image

jdneo:
fix #939
fix #938
fix #933

This PR include the following change
- Wrap the word for the stacktrace area
- Add ellipsis style to The class and method name
- Change the time format from `ms` to `s`

Kapture 2020-02-21 at 18 53 46

### DIFF
--- a/resources/templates/report_method_table.pug
+++ b/resources/templates/report_method_table.pug
@@ -5,16 +5,16 @@ mixin collapseMethodTable(children, type)
         - classIdx++
         ul.list-group.mb-3
             li.list-group-item.active
-                h5.mb-0 #{entry[0]}
+                h5.mb-0.text-truncate-left #{entry[0]}
             - var methodIdx = 0
             each method in entry[1]
                 - methodIdx++
                 li.list-group-item.list-group-item-action
                     div.row.accordion-toggle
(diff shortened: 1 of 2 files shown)


jdneo:
> LGTM, but why swtich to `s`?

It's to keep the same time unit as Eclipse

spottedmahn:
thanks @jdneo!! 🤝😀

PR #946, opened 2020-02-24
Fix the regression bug: assume failure is not marked as skipped

### LINKED ISSUE #945: Test runner does not skip assume case when the condition's not met
### Environment:
OS: Win10
JDK: 8
Java Test Runner Extension: 0.22.2

### Repro steps:

1. Git clone http://www.mkyong.com/wp-content/uploads/2014/01/TestNG-Spring-Example.zip
2. Add a Assume test case, like this:
```
import static org.junit.Assert.assertEquals;
import org.junit.Assume;
import org.junit.Ignore;
import org.junit.Test;

public class MyAssumeTest {
    
    @Test
    public void test() {
        Assume.assumeTrue(false);
        assertEquals(3, 1+2);
    }

    @Test
    @Ignore
    public void test1() {
        throw new RuntimeException();
    }
}
```
3. Run the MyAssumeTest java file

### Result:
The test does not skip this case
image

### Expect:
image

jdneo:
fix #945

### DIFF
--- a/src/runners/junitRunner/JUnitRunnerResultAnalyzer.ts
+++ b/src/runners/junitRunner/JUnitRunnerResultAnalyzer.ts
@@ -70,12 +70,8 @@ export class JUnitRunnerResultAnalyzer extends BaseRunnerResultAnalyzer {
                 this.currentTestItem = testId;
                 const failedResult: ITestResult = Object.assign({}, testResultManager.getResultById(testId), {
                     id: testId,
-                    status: TestStatus.Fail,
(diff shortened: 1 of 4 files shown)


jdneo:
It's ok whether the assume statement is in the first line or not, the Test Runner will handle that situation and return back the right output to the output parser.

The Assume is used to do some condition check, so you can put wherever you want as long as you want to do some condition check. For example, if I want to make sure the test code will only be executed in the Windows environment, then I can write: `AssumeTrue(System.getProperty("os.name").startsWith("Windows"))`

PR #997, opened 2020-05-18
Only resolve the code lens when then opened file is on test source path

jdneo:


src/testFileWatcher.ts
```diff
@@ -11,15 +13,47 @@ import { getTestSourcePaths } from './utils/commandUtils';
 
 class TestFileWatcher implements Disposable {
 
+    private testSourcePaths: string[] = [];
     private disposables: Disposable[] = [];
+    private registerListenersDebounce: (() => Promise<void>) & _.Cancelable = _.debounce(this.registerListenersInternal, 2 * 1000 /*ms*/);
 
-    public async registerListeners(): Promise<void> {
+    public async registerListeners(debounce: boolean = true): Promise<void> {
```
testforstephen, line 20: how about change the default to false? so if you want to register directly, then just call registerListeners(), that looks more natural to me.

jdneo: ok

src/codelens/TestCodeLensProvider.ts
```diff
@@ -21,6 +22,10 @@ export class TestCodeLensProvider implements CodeLensProvider, Disposable {
     }
 
     public async provideCodeLenses(document: TextDocument, _token: CancellationToken): Promise<CodeLens[]> {
+        if (!testFileWatcher.isOnTestSourcePath(document.uri.fsPath)) {
```
testforstephen, line 25: About filter the codelens, it seems vscode supports glob pattern on whether to trigger codelens.

```typescript
	export interface DocumentFilter {

		/**
		 * A language id, like `typescript`.
		 */
		language?: string;

		/**
		 * A Uri [scheme](#Uri.scheme), like `file` or `untitled`.
		 */
		scheme?: string;

		/**
		 * A [glob pattern](#GlobPattern) that is matched on the absolute path of the document. Use a [relative pattern](#RelativePattern)
		 * to filter documents to a [workspace folder](#WorkspaceFolder).
		 */
		pattern?: GlobPattern;
	}

```

PR #1019, opened 2020-06-22
feat: Adopt the Java Language Server's new API

jdneo:
demo

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,44 +74,44 @@
             "view/title": [
                 {
                     "command": "java.test.explorer.runAll",
-                    "when": "view == testExplorer",
+                    "when": "view == testExplorer && java:serverMode != LightWeight",
                     "group": "navigation@0"
                 },
                 {
                     "command": "java.test.explorer.debugAll",
-                    "when": "view == testExplorer",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -74,44 +74,44 @@
             "view/title": [
                 {
                     "command": "java.test.explorer.runAll",
-                    "when": "view == testExplorer",
+                    "when": "view == testExplorer && serverMode != LightWeight",
```
Eskibear, line 77: As we dicussed before, there's potential risk that the name of this context value, `serverMode`,  conflicts with other extensions or vscode itself. And here it seems to depend on redhat.java to set this context value.

As it's available to fetch `serverMode` thru extension.api, I suggest to set context value in this extension,  with some identifier like `java-test:serverMode`, just as GitLens is doing.
But the cons are:
* other extensions also have to define duplicate context values.
* the name is confusing, indicating that `serverMode` belongs to test runner (but not true).

jdneo: Or rename the `serverMode` to `java:serverMode` in the upstream once the confliction occurs?

jdneo: Wait for https://github.com/redhat-developer/vscode-java/pull/1504

Issue #1014, opened 2020-06-08
No code lens in version 0.23.0


ecmel:
Code lens is not working in 0.23.0 version.

jdneo:
@ecmel We did a change in 0.23.0, which will not resolve Code Lens in non-test source path.

Could you check if you are editing source files under test source path? Or could you please share a sample project to us?

ecmel:
@jdneo I created the project using maven-archetype-quickstart. JUnit 4 test is under src/test/java/{package}

jdneo:
I cannot repro on my side:

![image](https://user-images.githubusercontent.com/6193897/84107489-5649bc80-aa50-11ea-8d25-7416f28d2494.png)

Please check:
1. Is the language server running in `LightWeight` mode? -- Check the setting `java.server.launchMode`
2. Is the language server working: You can test this by simply run the `main` method in src/main/java/{package}/App.java

ecmel:
1. Language server is running in default hybrid mode.
2. Main method can be executed.
![Capture1](https://user-images.githubusercontent.com/1858678/84108915-1079fd80-aa2a-11ea-930b-1b3aa3878246.PNG)

Code lens disappeared after 0.23.0 upgrade:
![Capture2](https://user-images.githubusercontent.com/1858678/84108969-36070700-aa2a-11ea-97ab-e21a98120528.PNG)

jdneo:
I see. I think the major difference between our usage is: I'm opening the folder locally but you are in WSL.

I also noticed that in your file explorer, there is no `.classpath` & `.project` files. Did you hide them from the explorer?

ecmel:
Yes .classpath and .project are hidden and I am on wsl

jdneo:
@ecmel would you mind to share the content of the `.classpath` file?

ecmel:
`
<?xml version="1.0" encoding="UTF-8"?>
<classpath>
	<classpathentry kind="src" output="target/classes" path="src/main/java">
		<attributes>
			<attribute name="optional" value="true"/>
			<attribute name="maven.pomderived" value="true"/>
		</attributes>
	</classpathentry>
	<classpathentry kind="src" output="target/test-classes" path="src/test/java">
		<attributes>
			<attribute name="optional" value="true"/>
			<attribute name="maven.pomderived" value="true"/>
			<attribute name="test" value="true"/>
		</attributes>
	</classpathentry>
[...]

PR #1021, opened 2020-06-27
set document selector with pattern string

nikhilogic:
Fix for issue https://github.com/microsoft/vscode-java-test/issues/1014
The  languages.registerCodeLensProvider(documentSelector... ) api is not able to handle the RelativePattern object in the DocumentFilter object. This was breaking the code lens.
It is replaced by String pattern.


jdneo:
Thank you @nikhilogic for your contribution.

This is actually a bug of upstream VS Code. I've created https://github.com/microsoft/vscode/issues/99938 and it has been fixed. So I think there is no need to change the implementation for the Test Runner side.

Issue #531, opened 2018-12-18
Keyboard shortcut


franciscocpg:
I was able to configure a keyboard shortcut to run **all project tests** with the following configuration at `keybindings.json`.
```json
(...)
 {
        "key": "ctrl+t",
        "command": "java.test.explorer.run",
        "when": "resourceLangId == java"
 }
(...)
```
But I want to be able to configure 2 more shortcuts to:
1. Run all tests from the current file
2. Run test from the current focused method

Is there a way to do so?

jdneo:
@franciscocpg Unfortunately, there is no way to do this for now.

We will consider adding this support in the future.

tkleczek:
Would be also nice to be able to configure shortcut to re-run last "Run test" command

JasonHenriksen:
> Would be also nice to be able to configure shortcut to re-run last "Run test" command

I'm considering writing a PR to do exactly this.  Would the Test Runner team be open to a PR like that?

jdneo:
@JasonHenriksen All the contributions are welcome! Please go ahead!

jdneo:
Hi @JasonHenriksen, is everything works fine? Is there anything I can help with?

jdneo:
Some new commands are introduced in the next release version 0.24.0, which you can set shortcuts for them:

- Open the test report: `java.test.show.report`
- Relaunch the test execution: `java.test.relaunch`

baldimir:
Hi, any idea when this could be implemented? I see it is in backlog, just asking if you are at least considering this in near future.

jdneo:
@baldimir The next release will have a new command to run the test in the current editing file. See: #1066

baldimir:
Thanks, so what is left is a command to run a specific test from a file.

jdneo:
New commands added in the `0.25.0` to run/debug the tests in current file:

- java.test.editor.run
- java.test.editor.debug

jdneo:
In 0.31.0, we support the newest VS Code testing API in 1.59.0.

There are some new commands added, which are useful and cover the request of this issue, please have a try!

![image](https://user-images.githubusercontent.com/6193897/128446850-cb8d293e-f5d3-4fbe-bda4-c7c78153dc07.png)

jscherer:
Love this :-)
Maybe someone should update the (https://code.visualstudio.com/docs/java/java-testing) documentation with a keyboard shortcut section, because it took me quite a while to find this info!

PR #1037, opened 2020-07-18
feat: run all tests from active editor

anothernode:
Hi!

I managed to create a command that runs all tests from the currently opened file in the active editor, thus addressing the first part of #531.

I'm completely new to VS Code Extension development and almost completely new to JavaScript/TypeScript and I'm pretty sure there's a lot of room for improvement in terms of code style and robustness. So if you folks see any problems with the code I'd be happy to revise the PR according to your guidance :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
         "onCommand:java.test.explorer.refresh",
         "onCommand:java.test.explorer.run",
         "onCommand:java.test.explorer.debug",
+        "onCommand:java.test.activeEditor.run",
         "onCommand:java.test.run",
         "onCommand:java.test.debug",
         "onCommand:java.test.cancel",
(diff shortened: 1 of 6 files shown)


anothernode:
I don't understand why the checks are failing. When I run `npm run test` locally, I don't see any failing tests.

Let me know if there's anything I can do to help getting the checks to pass.

jdneo:
Thank you @anothernode, 

I'm thinking that if we can use just one new command to achieve all, for example:

1. if focus is a java file/folder in file explorer, run all the test in the classes/packages -> similar to run test from test explorer
2. if focus is in editor, and the cursor belongs to a class scope, run the tests in the class.
3. if focus is in editor, and the cursor belongs to a method scope, run this test method.

The 2 & 3 might be able to achieved by get the covering AST node, which could be queried from the Java Language Server. Let me think more about the solution, and update to you later.

anothernode:
Thanks for your feedback @jdneo!

I agree that your suggestion would be a more comprehensive and nicer solution. If I find the time I'll try and work out the details about how to query the Java Language Server for figuring out in which scope the cursor currently is and how to make the distinction between 1) focus in explorer and 2/3) focus in editor.

If you can give any further hints on how to approach this, example code, docs etc. I'd very much appreciate any advice.

anothernode:
Thankfully this was implemented in a nicer way by @jdneo

jdneo:
@anothernode Thank you for your contribution.

PR #1051, opened 2020-08-31
Use adaptive debounce when resolving code lenses

jdneo:


src/codelens/TestCodeLensProvider.ts
```diff
@@ -26,14 +30,25 @@ export class TestCodeLensProvider implements CodeLensProvider, Disposable {
...
 
         try {
-            const items: ITestItem[] = await testItemModel.getItemsForCodeLens(document.uri);
-            return this.getCodeLenses(items);
+            const timeout: number = this.getRequestDelay(document.uri);
+            return new Promise<CodeLens[]>((resolve: (ids: CodeLens[]) => void): void => {
+                token.onCancellationRequested(() => {
+                    clearTimeout(timeoutHandle);
+                    resolve([]);
+                    this.dispose();
```
testforstephen, line 44: what's the purpose to dispose here?

jdneo: Yes this seems can be removed. (Checked at the VS Code repo about the usage of this, and not found such kind of usage)

testforstephen (approved):

PR #1052, opened 2020-08-31
Get ASTNode from ASTProvider

jdneo:
Did some profiling to check the improvement. Given a 5000+ lines test file, doing the following operations:
- 5 times of code completion (keep enter `.` and `backspace`)
- write `String a = "";`
- write `String b = a;`
- insert code snippet `systrace`.

### Before
WeChat Screenshot_20200831144332

### After
WeChat Screenshot_20200831143822

### DIFF
--- a/java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestSearchUtils.java
+++ b/java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestSearchUtils.java
@@ -41,6 +41,7 @@ import org.eclipse.jdt.core.dom.ASTParser;
 import org.eclipse.jdt.core.dom.CompilationUnit;
 import org.eclipse.jdt.core.dom.ITypeBinding;
 import org.eclipse.jdt.core.dom.TypeDeclaration;
+import org.eclipse.jdt.core.manipulation.CoreASTProvider;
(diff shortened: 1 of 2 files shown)


java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestSearchUtils.java
```diff
@@ -111,12 +112,16 @@
     }
 
     public static ASTNode parseToAst(final ICompilationUnit unit, IProgressMonitor monitor) {
-        final ASTParser parser = ASTParser.newParser(AST.JLS14);
-        parser.setSource(unit);
-        parser.setFocalPosition(0);
-        parser.setResolveBindings(true);
-        parser.setIgnoreMethodBodies(true);
-        return parser.createAST(monitor);
+        final CompilationUnit astRoot = CoreASTProvider.getInstance().getAST(unit, CoreASTProvider.WAIT_YES, monitor);
```
testforstephen, line 115: is there any difference if setting to WAIT_NO?

jdneo: Check and not very significant difference observed

testforstephen (approved):

PR #1066, opened 2020-09-29
feat: Support run tests from current file

jdneo:
Add a new command to support run tests in current active file.

demo

Part of #531

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,8 @@
         "onCommand:java.test.explorer.refresh",
         "onCommand:java.test.explorer.run",
         "onCommand:java.test.explorer.debug",
+        "onCommand:java.test.editor.run",
+        "onCommand:java.test.editor.debug",
         "onCommand:java.test.run",
         "onCommand:java.test.debug",
         "onCommand:java.test.cancel",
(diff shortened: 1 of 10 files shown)


Eskibear:
what if I call the command in a .java file outside the workspace? E.g. I drag one file from another project, opening it in current window.

package.json
```diff
@@ -203,6 +205,16 @@
...
+            {
+                "command": "java.test.editor.run",
+                "title": "%contributes.commands.java.test.editor.run.title%",
+                "category": "Java"
+            },
+            {
+                "command": "java.test.editor.debug",
+                "title": "%contributes.commands.java.test.editor.debug.title%",
+                "category": "Java"
+            },
```
testforstephen, line 217: You only define a command in this PR, why not to expose it to the editor context menu either?

jdneo: Per offline discussion. We will expose these commands into editor context menu (through new submenu contribution point) and editor title in the next release (0.26.0).

Eskibear (approved):

testforstephen (approved):

PR #1084, opened 2020-10-20
refactor: Only show migrate command when workspace has deprecated files

jdneo:


src/extension.ts
```diff
@@ -127,6 +130,14 @@ export function isSwitchingServer(): boolean {
     return serverMode === LanguageServerMode.Hybrid;
 }
 
+async function setContextKeyForDeprecatedConfig(): Promise<void> {
+    const deprecatedConfigs: Uri[] = await workspace.findFiles('**/.vscode/launch.test.json', null, 1 /*maxResult*/);
+    if (deprecatedConfigs.length > 0) {
+        commands.executeCommand('setContext', 'java:hasDeprecatedTestConfig', true);
+        sendInfo('', { hasDeprecatedTestConfig: 1 });
```
Eskibear, line 137: ` hasDeprecatedTestConfig: 1` would cause telemetry data goes into `measure` instead of `properties`.    Consider to use `true` or `"1"`?

Eskibear (approved):

PR #1085, opened 2020-10-20
feat: Add shortcut buttons in editor toolbar

jdneo:
Show the `run test` button in editor toolbar if the java file locates in `src/test/java` and ends with `T(t)est(s)`

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,6 +80,18 @@
             }
         ],
         "menus": {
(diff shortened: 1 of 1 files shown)


Eskibear (approved):

jdneo:
Per offline discussion, it will be handled by Java Debugger

Issue #1087, opened 2020-10-29
cannot read property 'error' of null


hacke2:
If execute `testFileWatcher.registerListeners` for the first time, it could throw error but logger is not initialized.

extension.ts:

```ts
await testFileWatcher.registerListeners();
testExplorer.initialize(context);
runnerScheduler.initialize(context);
testReportProvider.initialize(context);

const storagePath: string = context.storagePath || path.join(os.tmpdir(), 'java_test_runner');
await fse.ensureDir(storagePath);
logger.initialize(storagePath, context.subscriptions);
```

testFileWatcher.ts:

```ts
try {
  const sourcePaths: string[] = await getTestSourcePaths(workspace.workspaceFolders.map((workspaceFolder: WorkspaceFolder) => workspaceFolder.uri.toString()));
  for (const sourcePath of sourcePaths) {
    const normalizedPath: string = Uri.file(sourcePath).fsPath;
    const pattern: RelativePattern = new RelativePattern(normalizedPath, '**/*.java');
    this.patterns.push(pattern);
    const watcher: FileSystemWatcher = workspace.createFileSystemWatcher(pattern, true /* ignoreCreateEvents */);
    this.registerWatcherListeners(watcher);
    this.disposables.push(watcher);
  }
} catch (error) {
    logger.error('Failed to get the test paths', error);
    const watcher: FileSystemWatcher = workspace.createFileSystemWatcher('**/*.java');
    this.registerWatcherListeners(watcher);
    this.disposables.push(watcher);
}
```

PR #1088, opened 2020-10-29
fix: Initialize logger ahead of time

hacke2:
Initialize logger ahead of time

Fixed: #1087


jdneo (approved):

jdneo:
Thank you @hacke2 for reporting the bug and providing the fix!

PR #1089, opened 2020-11-03
feat: Navigate from stack trace to source

### LINKED ISSUE #384: Navigate from stacktrace line to proper file in test results
Following up on #306, we need better navigation from the report to java files. Since this is gonna be more complicated I'm opening it in a separate ticket: I'd be nice if the generated stacktraces actually linked to matching lines from source code, to make it possible to navigate directly from the report to any file, in the workspace or the classpath, involved in the test failure, as with other Java IDEs.

jdneo:
resolve #384

demo

### DIFF
--- a/src/utils/commandUtils.ts
+++ b/src/utils/commandUtils.ts
@@ -36,6 +36,11 @@ export async function searchTestLocation(fullName: string): Promise<ILocation[]>
         JavaTestRunnerDelegateCommands.SEARCH_TEST_LOCATION, fullName) || [];
 }
 
+export async function resolveSourceUri(trace: string, projectNames: string[]): Promise<string> {
+    return await executeJavaLanguageServerCommand<string>(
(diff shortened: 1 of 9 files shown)


src/utils/commandUtils.ts
```diff
@@ -36,6 +36,11 @@ export async function searchTestLocation(fullName: string): Promise<ILocation[]>
         JavaTestRunnerDelegateCommands.SEARCH_TEST_LOCATION, fullName) || [];
 }
 
+export async function resolveSourceUri(trace: string, projectNames: string[]): Promise<string> {
```
testforstephen, line 39: why not change the name to resolveStackTraceLocation directly?

jdneo: ok

testforstephen (approved):

PR #1112, opened 2020-12-01
fix: Open debug console on session start

### LINKED ISSUE #1106: Do not force the execution inside the debug console
Some users may expect that the output should be seen in the integrated terminal (https://github.com/microsoft/vscode-java-test/issues/344#issuecomment-731977823), while it's actually in the debug console (which is forced by this extension). So we should consider to follow the same behavior as the Java Debugger.

jdneo:
fix #1106


jdneo:
test failed because the command argument might have `(` and `)` in the method signature, need to find a way to escape them

testforstephen (approved):

PR #1127, opened 2021-01-13
fix: Add label information to the test result status

### LINKED ISSUE #1126: Please provide custom aria labels for status bar items.
CC: @isidorn

As I am blind, I use VSCode with the help of a screen reader to execute my test cases. After running the tests, I access the status bar to check the test results.
The only information present in the status bar that is apparently related to the test results and which is presented to me by the screen reader is the following:
'x 0 check 3'

It seems to me that the extension could make use of aria labels with information about the results. Screen readers would pass this information on to users without any major problems.

Actually I am running code-insiders 1.53.0-insider in my arch linux box.

jdneo:
fix #1126

### DIFF
--- a/src/testStatusBarProvider.ts
+++ b/src/testStatusBarProvider.ts
@@ -41,7 +41,7 @@ class TestStatusBarProvider implements Disposable {
         }
 
         this.statusBarItem.accessibilityInformation = {
-            label: `${failedNum} failed, ${passedNum} passed test${failedNum + passedNum > 1 ? 's' : ''}`,
+            label: `${failedNum} failed, ${passedNum} passed test${failedNum + passedNum === 1 ? '' : 's'}`,
         };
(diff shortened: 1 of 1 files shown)


jdneo:
@isidorn This is the PR to resolve #1126. Please also take a look when you have time, thanks!

Eskibear (dismissed):

src/testStatusBarProvider.ts
```diff
@@ -40,6 +40,9 @@ class TestStatusBarProvider implements Disposable {
             }
         }
 
+        this.statusBarItem.accessibilityInformation = {
+            label: `${failedNum} failed, ${passedNum} passed test${failedNum + passedNum > 1 ? 's' : ''}`,
```
Eskibear, line 44: AFAIK, 0 is plural in English, maybe `num === 1 ? '' : "s" ` would be better?

jdneo: yes that's better!

isidorn (approved):

isidorn:
Looks good!

Eskibear (approved):

jdneo:
Thank you @isidorn!

PR #1131, opened 2021-01-26
refactor: Leverage test finder for test discovery

### LINKED ISSUE #641: The test cases will run multiple times.
If the user triggers tests from the workspace level, and if the user has some classes annotated with `@Suite`, then some of the test cases will run multiple time.

### LINKED ISSUE #773: Cannot run meta annotation tests from test explorer

### LINKED ISSUE #818: Support for method detection with @Testable (and meta)
In #683 it was mentioned that `@Testable` is designed for support in IDEs (so exactly the use case that `vscode-java-test` was created for), but there is no support yet. It would be really useful, especially if supported (as intended according to the javadocs) as a meta annotation.

jdneo:
This PR leverage the upstream Test Finder for test discovery.

The previous implementation used `SearchEngine`, since it's not good at handleing meta annotation, some bugs are introduced.

fix #641
fix #773
fix #818

### DIFF
--- a/java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/searcher/JUnit4TestSearcher.java
+++ b/java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/searcher/JUnit4TestSearcher.java
@@ -11,24 +11,35 @@
 
 package com.microsoft.java.test.plugin.searcher;
 
+import com.microsoft.java.test.plugin.model.TestItem;
 import com.microsoft.java.test.plugin.model.TestKind;
-import com.microsoft.java.test.plugin.util.TestFrameworkUtils;
+import com.microsoft.java.test.plugin.model.TestLevel;
+import com.microsoft.java.test.plugin.util.TestItemUtils;
 
-import org.eclipse.jdt.core.Flags;
-import org.eclipse.jdt.core.IMethod;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.jdt.core.IJavaElement;
+import org.eclipse.jdt.core.IType;
 import org.eclipse.jdt.core.JavaModelException;
 import org.eclipse.jdt.core.dom.IMethodBinding;
 import org.eclipse.jdt.core.dom.Modifier;
+import org.eclipse.jdt.internal.junit.launcher.JUnit4TestFinder;
 import org.eclipse.jdt.internal.junit.launcher.TestKindRegistry;
+
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 public class JUnit4TestSearcher extends BaseFrameworkSearcher {
 
-    public static final String RUN_WITH = "org.junit.runner.RunWith";
+    private static final JUnit4TestFinder JUNIT4_TEST_FINDER = new JUnit4TestFinder();
 
     public JUnit4TestSearcher() {
         super();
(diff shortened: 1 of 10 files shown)


java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestSearchUtils.java
```diff
@@ -189,63 +193,66 @@ public static ASTNode parseToAst(final ICompilationUnit unit, IProgressMonitor m
         final Gson gson = new Gson();
         final SearchTestItemParams params = gson.fromJson((String) arguments.get(0), SearchTestItemParams.class);
 
-        final IJavaSearchScope scope = createSearchScope(params);
+        if (params.getLevel() == TestLevel.METHOD) {
+            // unreachable code since the client will directly run the test when it's triggered from method level
+            return Collections.emptyList();
```
Eskibear, line 198: if it's unreachable, maybe throw excepetion or do assertion? (when unexpected scenario does occur, it would be fast located..)

java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/searcher/JUnit5TestSearcher.java
```diff
@@ -151,6 +114,29 @@ public boolean findAnnotation(IAnnotationBinding[] annotations, String[] annotat
...
+
+    @Override
+    public TestItem[] findTestsInContainer(IJavaElement element, IProgressMonitor monitor) throws CoreException {
+        final Map<String, TestItem> result = new HashMap<>();
+        final Set<IType> types = new HashSet<>();
+        JUNIT5_TEST_FINDER.findTestsInContainer(element, types, monitor);
+        for (final IType type : types) {
+            final TestItem item = TestItemUtils.constructTestItem(type, TestLevel.CLASS, TestKind.JUnit5);
+            item.setChildren(Arrays.stream(type.getMethods())
+                    .map(m -> String.format("%s@%s", m.getJavaProject().getProject().getName(),
```
Eskibear, line 130: `String.format("%s@%s"` is much slower. Because the pattern is simple, suggest to use  `string + "@" + string`, which will be optimized by JVM to use StringBuilder under the hood.

java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/searcher/JUnit4TestSearcher.java
```diff
@@ -77,4 +65,27 @@ public boolean isTestMethod(IMethodBinding methodBinding) {
...
+
+    @Override
+    public TestItem[] findTestsInContainer(IJavaElement element, IProgressMonitor monitor) throws CoreException {
+        final Map<String, TestItem> result = new HashMap<>();
+        final Set<IType> types = new HashSet<>();
+        JUNIT4_TEST_FINDER.findTestsInContainer(element, types, monitor);
+        for (final IType type : types) {
+            final TestItem item = TestItemUtils.constructTestItem(type, TestLevel.CLASS, TestKind.JUnit);
+            item.setChildren(Arrays.stream(type.getMethods())
+                    .map(m -> String.format("%s@%s", m.getJavaProject().getProject().getName(),
```
Eskibear, line 82: String.format is slower

Eskibear (approved):

PR #1137, opened 2021-02-01
chore: Use GitHub Actions for CI

jdneo:


src/commands/testReportCommands.ts
```diff
@@ -39,7 +39,8 @@ export async function openTestSourceLocation(uri: string, range: string, fullNam
     if (uri && range) {
         return commands.executeCommand(JavaTestRunnerCommands.OPEN_DOCUMENT, Uri.parse(uri), JSON.parse(range) as Range);
     } else if (fullName) {
-        const items: ILocation[] = await searchTestLocation(fullName.slice(fullName.indexOf('@') + 1, fullName.indexOf('[')));
+        const methodEndIndex: number = fullName.indexOf('[');
+        const items: ILocation[] = await searchTestLocation(fullName.slice(fullName.indexOf('@') + 1, methodEndIndex < 0 ? undefined : methodEndIndex));
```
testforstephen, line 43: what's the purpose of this change?

jdneo: I should only provide the `end` param when the string has `'['`, otherwise `indexOf('[')` will return -1 thus the last character will be truncated. 

It's a regression introduced in #1135. It will actually fail the CI, but since tracis.org is not working so I didn't find it in time.

testforstephen:
@majguo could you help review whether there is security vulnerability?

majguo (approved):

testforstephen (approved):
LGTM

Issue #1172, opened 2021-04-12
Support Generate test


jdneo:
This include two different features

- [x] Add test method and lifecycle methods from a test source
- [x] Generate test from a source (scaffolding)

PR #1173, opened 2021-04-13
feat: Support generate test methods in a test Java file

jdneo:
Address the first part of #1172  -- Add test method and lifecycle methods from a test source

### UI entry
`Source Actions` > `Generate Tests...`

### Steps

1. determine the test framework
    - check from the import statement
    - if no test related imports found, check classpath
2. determine the methods to insert
    - Available options are:
        - Test
        - 4 lifecycle methods (setUp, tearDown, beforeClass, afterClass)
    - Existing lifecycle methods will be filtered out
3. Create the workspace edit and apply


java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestGenerationUtils.java
```diff
@@ -0,0 +1,407 @@
...
+
+        final TextEdit textEdit = createTextEdit(testKind, methodsToGenerate, root, typeNode,
+                typeBinding, insertPosition);
+        return SourceAssistProcessor.convertToWorkspaceEdit(unit, textEdit);
+    }
+
+    private static List<String> determineMethodsToGenerate(List<String> lifecycleAnnotations) {
+        final List<String> methodList = lifecycleAnnotations.stream()
+                .map(annotation -> "@" + annotation.substring(annotation.lastIndexOf(".") + 1) + " Method")
+                .collect(Collectors.toList());
```
testforstephen, line 190: If the helper `getLifeCycleAnnotations()` just returns the simple annotation names without prefix, you could use lambda `annotation -> "@" + annotation + " Method"` directly.

so suggest to refactor  `getLifeCycleAnnotations()` to return the simple names instead of qualified names.

java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestGenerationUtils.java
```diff
@@ -0,0 +1,407 @@
...
+        edit.addChild(importRewrite.rewriteImports(null));
+        edit.addChild(astRewrite.rewriteAST());
+        return edit;
+    }
+
+    private static List<String> getLifecycleAnnotations(TestKind testKind) {
+        final List<String> list = new ArrayList<>();
+        switch (testKind) {
+            case JUnit:
+                list.add(JUNIT4_LIFECYCLE_ANNOTATION_PREFIX + JUNIT4_BEFORE_CLASS_ANNOTATION);
```
testforstephen, line 302: No need to include the prefix.

java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestGenerationUtils.java
```diff
@@ -0,0 +1,407 @@
...
+
+        return null;
+    }
+
+    private static void generateTestsFromSource() {
+        // TODO: unimplement
+    }
+
+    private static WorkspaceEdit generateTestsFromTest(ICompilationUnit unit, CompilationUnit root,
+            TypeDeclaration typeNode, ITypeBinding typeBinding, int cursorPosition)
```
testforstephen, line 155: cursorPosition -> cursorOffset

java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestGenerationUtils.java
```diff
@@ -0,0 +1,407 @@
...
+                list.add(TESTNG_LIFECYCLE_ANNOTATION_PREFIX + TESTNG_TEAR_DOWN_ANNOTATION);
+                list.add(TESTNG_LIFECYCLE_ANNOTATION_PREFIX + TESTNG_AFTER_CLASS_ANNOTATION);
+                break;
+            default:
+                break;
+        }
+        return list;
+    }
+
+    private static String getPromptNameByAnnotation(String annotation) {
```
testforstephen, line 325: -> getSuggestedMethodNameByAnnotation(String annotation)

src/codeActionProvider.ts
```diff
@@ -0,0 +1,50 @@
...
+import { CancellationToken, CodeAction, CodeActionContext, CodeActionKind, CodeActionProvider, Range, Selection, TextDocument } from 'vscode';
+import { getSourcePaths } from './utils/commandUtils';
+
+export class TestCodeActionProvider implements CodeActionProvider {
+
+    private sourcePaths: ISourcePath[];
+
+    public async provideCodeActions(document: TextDocument, range: Range | Selection, _context: CodeActionContext, token: CancellationToken): Promise<CodeAction[] | null> {
+        if (!this.sourcePaths) {
+            this.sourcePaths = await getSourcePaths();
```
testforstephen, line 14: No need to restrict it to files under source paths. If the source paths have changed, it won't work.

You can just enable the source action at any *.java file.

testforstephen (changes requested):

java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/model/TestKind.java
```diff
@@ -1,14 +1,79 @@
...
+    }
+
+    public int getValueForRank() {
+        switch (this) {
+            case JUnit:
+                return 1;
+            case JUnit5:
+                return 0;
+            case TestNG:
+                return 2;
```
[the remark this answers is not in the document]
jdneo: I intended to not change the value of the enum which implemented before(afraid of introducing any breaking 😄 ). But definitely, use value as the rank will make the code much simpler.

Will change it in the next commit

src/codeActionProvider.ts
jdneo: And will lessen the time to resolve the code actions. Makes sense.

java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestGenerationUtils.java
```diff
@@ -0,0 +1,407 @@
...
+                    "Cannot find a unit test framework in the project, please make sure it's on the classpath.");
+            return null;
+        }
+        if (availableFrameworks.size() == 1) {
+            return availableFrameworks.iterator().next();
+        } else {
+            final List<String> frameworkList = availableFrameworks.stream()
+                .sorted((kind1, kind2) -> kind1.getValueForRank() - kind2.getValueForRank())
+                .map(framework -> framework.toString())
+                .collect(Collectors.toList());
```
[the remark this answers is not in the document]
jdneo: Since `availableFrameworks` is set, so I think we still need sort it?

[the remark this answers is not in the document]
testforstephen: When you find available frameworks, you already follow the order [junit5, junit4, testng] to search them. This is already the expected order you want.

testforstephen (approved):

PR #1178, opened 2021-04-23
refactor: Decouple the codelens provider and test file watchers

jdneo:
- Add a `testSourceProvider` so that the `codelensProvider` and `testFileWatchers` can be decoupled and handle the initialization by themselves.


src/provider/testSourceProvider.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    public async initialize(): Promise<void> {
+        this.testSource = [];
+        this.strictTestSource = [];
+        if (!workspace.workspaceFolders) {
+            return;
+        }
+
+        this.testSource = await getTestSourcePaths(workspace.workspaceFolders.map((workspaceFolder: WorkspaceFolder) => workspaceFolder.uri.toString()), true);
+        this.strictTestSource = await getTestSourcePaths(workspace.workspaceFolders.map((workspaceFolder: WorkspaceFolder) => workspaceFolder.uri.toString()), false);
```
Eskibear, line 26: For simplicity it's ok. But actually these are uplicate calls, same check is executed on same workspaceFolders. Consider to refine the implementation if it's expensive, e.g. return both in one call, one check.

jdneo: make sense

Eskibear (approved):

PR #1181, opened 2021-04-25
chore: Only copy test files in development mode

jdneo:
Mark as draft for now. So there are several things needs to do:

- update webpack
- update package.json
- update GitHub actions definition


jdneo:
Close since the code base has been changed a lot.

PR #1183, opened 2021-04-26
fix: Generated test methods reduced the visibility of the inherited methods

### LINKED ISSUE #1182: Generated test methods reduced the visibility of the duplicated methods from inherited class
Below are samples.
```java
public class A {
    @BeforeAll
    public static void beforeClass() {
    }

    @BeforeEach
    public void setUp() {
    }
}
```

```java
public class B extends A {
    @BeforeAll
    static void beforeClass() { // Generated lifecycle method will report error "Cannot reduce the visibility of the inherited method from A".
    }

    @BeforeEach
    void setUp() { // Generated lifecycle method will report error "Cannot reduce the visibility of the inherited method from A".
    }
}
```

Should check if the inherited class contains the same methods. If they're visible to the subclass, the subclass should not reduce the visibility of these methods. Also add `@Override` annotation for them.

jdneo:
Fix #1182

### DIFF
--- a/java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestGenerationUtils.java
+++ b/java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestGenerationUtils.java
@@ -43,6 +43,7 @@ import org.eclipse.jdt.core.dom.rewrite.ImportRewrite;
 import org.eclipse.jdt.core.dom.rewrite.ImportRewrite.ImportRewriteContext;
 import org.eclipse.jdt.core.dom.rewrite.ListRewrite;
 import org.eclipse.jdt.internal.core.manipulation.StubUtility;
+import org.eclipse.jdt.internal.corext.codemanipulation.CodeGenerationSettings;
 import org.eclipse.jdt.internal.corext.codemanipulation.ContextSensitiveImportRewriteContext;
 import org.eclipse.jdt.internal.corext.codemanipulation.StubUtility2Core;
 import org.eclipse.jdt.internal.corext.dom.ASTNodeFactory;
@@ -56,9 +57,10 @@ import org.eclipse.text.edits.MalformedTreeException;
 import org.eclipse.text.edits.MultiTextEdit;
 import org.eclipse.text.edits.TextEdit;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Objects;
+import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -128,6 +130,12 @@ public class TestGenerationUtils {
         final ITypeBinding binding = ((TypeDeclaration) coveringNode).resolveBinding();
         if (binding == null) {
             JUnitPlugin.logError("Failed to resolve type binding from " + unit.getElementName());
+            return null;
+        }
+
+        if (!binding.isClass()) {
+            JavaLanguageServerPlugin.getInstance().getClientConnection().showNotificationMessage(MessageType.Error,
+                    "Cannot generate tests if it's not a Java class.");
             return null;
         }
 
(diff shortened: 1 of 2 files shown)


java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestGenerationUtils.java
```diff
@@ -358,6 +354,42 @@ private static String getSuggestedMethodNameByAnnotation(String annotation) {
...
+            modifiers |= Modifier.PUBLIC;
+        }
+
+        // @BeforeClass and @AfterClass in JUnit 4 & 5 needs static modifier
+        if (needStaticModifier(kind, annotation)) {
+            modifiers |= Modifier.STATIC;
+        }
+
+        ITypeBinding superClass = typeBinding.getSuperclass();
+        while (superClass != null) {
```
testforstephen, line 374: If the super class contains a duplicate method, it's better to add `@Override` tag to the generated method in subclass.

java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestGenerationUtils.java
```diff
@@ -358,6 +354,42 @@ private static String getSuggestedMethodNameByAnnotation(String annotation) {
...
+            final IMethodBinding[] superMethods = superClass.getDeclaredMethods();
+            for (final IMethodBinding superMethod : superMethods) {
+                if (Objects.equals(superMethod.getName(), methodName)) {
+                    final int superModifiers = superMethod.getModifiers();
+                    // Cannot reduce the visibility of the inherited method
+                    if (Modifier.isPublic(superModifiers)) {
```
testforstephen, line 380: both `protected`, `public` methods in super class are visible to subclass, and could override it in subclass.

java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestGenerationUtils.java
```diff
@@ -264,19 +266,13 @@ private static TextEdit getTextEdit(TestKind kind, List<MethodMetaData> methodMe
...
-
-            // @BeforeClass and @AfterClass in JUnit 4 & 5 needs static modifier
-            if (needStaticModifier(kind, method.annotation)) {
-                decl.modifiers().addAll(ASTNodeFactory.newModifiers(ast, Modifier.STATIC));
-            }
 
             // set a unique method name according to the annotation type
-            decl.setName(ast.newSimpleName(getUniqueMethodName(typeBinding.getJavaElement(),
-                    method.methodName)));
+            final String methodName = getUniqueMethodName(typeBinding.getJavaElement(), method.methodName);
```
testforstephen, line 271: When finding unique method names, should also filter out the method names in super class with different signature.

java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestGenerationUtils.java
```diff
@@ -359,6 +373,60 @@ private static String getSuggestedMethodNameByAnnotation(String annotation) {
...
+            methods.put(methodBinding.getName(), methodBinding);
+        }
+        ITypeBinding superClass = typeBinding.getSuperclass();
+        while (superClass != null) {
+            for (final IMethodBinding methodBinding : superClass.getDeclaredMethods()) {
+                if (methods.containsKey(methodBinding.getName())) {
+                    continue;
+                }
+
+                if (Modifier.isPrivate(methodBinding.getModifiers())) {
```
testforstephen, line 420: https://docs.oracle.com/javase/tutorial/java/javaOO/accesscontrol.html

Members with private or no modifier are not accessible to subclass. Maybe the condition could be

```java
if (Modifier.isProtected(...) || Modifier.isPublic(...)) {
  methods.put(methodBinding.getName(), methodBinding);
}
```

java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestGenerationUtils.java
```diff
@@ -359,6 +374,77 @@ private static String getSuggestedMethodNameByAnnotation(String annotation) {
...
+        return methods;
+    }
+
+    private static boolean isAccessible(IMethodBinding superMethod, ITypeBinding declaredType) {
+        final int modifiers = superMethod.getModifiers();
+        if (Modifier.isPrivate(modifiers)) {
+            return false;
+        }
+
+        if (!Modifier.isProtected(modifiers) && !Modifier.isProtected(modifiers)) {
```
testforstephen, line 438: typos.

!Modifier.isPublic(...)

java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestGenerationUtils.java
```diff
@@ -359,6 +374,77 @@ private static String getSuggestedMethodNameByAnnotation(String annotation) {
...
+        if (binding == null) {
+            return modifiers;
+        }
+
+        final int superModifiers = binding.getModifiers();
+        if (Modifier.isProtected(superModifiers)) {
+            modifiers |= Modifier.PROTECTED;
+        } else if (Modifier.isPublic(superModifiers)) {
+            modifiers |= Modifier.PUBLIC;
+        }
```
testforstephen, line 403: if it's JUnit 4, always return public regardless of super implementation.

testforstephen (approved):

PR #1198, opened 2021-05-10
feat: Support select inherit methods when generating tests

jdneo:
part of #1172

This is to add the capability to select inherited methods. See:

https://user-images.githubusercontent.com/6193897/117597498-48c51700-b178-11eb-97b2-ef055b08dab5.mp4

### DIFF
--- a/java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/model/Option.java
+++ b/java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/model/Option.java
@@ -15,6 +15,7 @@ public class Option {
     public String value;
     public String label;
     public String description;
+    public boolean isAdvance;
 
     public Option(String label) {
         this.label = label;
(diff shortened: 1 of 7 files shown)


java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/model/Option.java
```diff
@@ -25,4 +26,9 @@ public Option(String value, String label, String description) {
         this.label = label;
         this.description = description;
     }
+
+    public Option(String value, String label, String description, boolean isAdvance) {
+        this(value, label, description);
+        this.isAdvance = isAdvance;
```
Eskibear, line 32: advanced？

Eskibear:
The UX LGTM, the only concern is "visibility" of the button as we discussed. I wonder how many users will notice it. 😆

testforstephen (dismissed):

Eskibear (approved):

PR #1200, opened 2021-05-12
disable extension in virtual workspaces

Eskibear:
see #1195

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,9 @@
       "request": "onDemand",
       "description": "Extension may require your trust on workspace for code execution."
     },
+    "capabilities": {
+      "virtualWorkspaces": false
+    },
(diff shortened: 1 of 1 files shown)


jdneo (approved):

package.json
```diff
@@ -31,6 +31,9 @@
       "request": "onDemand",
       "description": "Extension may require your trust on workspace for code execution."
     },
+    "capabilities": {
```
jdneo, line 34: BTW could you update the indentation to 4 spaces? including the above `workspaceTrust`

Eskibear: `workspaceTrust` part is out-of-date, I'll update it in a separate PR.

jdneo (approved):

PR #1207, opened 2021-05-24
fix: Can generate tests from interface, enum and record

### LINKED ISSUE #1206: Support "Generate Test..." in interface methods
image

Since the interfaces could add default implementation for method, we should allow to add tests for interface methods as well.

jdneo:
fix #1206

### DIFF
--- a/java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestGenerationUtils.java
+++ b/java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestGenerationUtils.java
@@ -126,7 +126,7 @@ public class TestGenerationUtils {
         ASTNode coveringNode = nodeFinder.getCoveringNode();
 
         while (coveringNode != null) {
-            if (coveringNode instanceof TypeDeclaration) {
+            if (coveringNode instanceof AbstractTypeDeclaration) {
                 break;
             }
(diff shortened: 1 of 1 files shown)


java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestGenerationUtils.java
```diff
@@ -139,20 +139,20 @@ public static WorkspaceEdit generateTests(List<Object> arguments, IProgressMonit
...
 
-        final ITypeBinding binding = ((TypeDeclaration) coveringNode).resolveBinding();
+        final ITypeBinding binding = ((AbstractTypeDeclaration) coveringNode).resolveBinding();
         if (binding == null) {
             JUnitPlugin.logError("Failed to resolve type binding from " + unit.getElementName());
             return null;
         }
 
-        if (!binding.isClass()) {
+        if (!(binding.isClass() || binding.isInterface() || binding.isRecord())) {
```
testforstephen, line 153: does this support enum type?

jdneo: Yes it does in IntelliJ, will cover that case

testforstephen (dismissed):

testforstephen (approved):

PR #1237, opened 2021-07-27
initialize the repo

jdneo:


java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/handler/TestDelegateCommandHandler.java
```diff
@@ -26,10 +26,6 @@
 public class TestDelegateCommandHandler implements IDelegateCommandHandler {
 
     private static final String GET_TEST_SOURCE_PATH = "vscode.java.test.get.testpath";
-    private static final String SEARCH_TEST_ITEMS = "vscode.java.test.search.items";
-    private static final String SEARCH_TEST_ALL_ITEMS = "vscode.java.test.search.items.all";
-    private static final String SEARCH_TEST_CODE_LENS = "vscode.java.test.search.codelens";
```
mfussenegger, line 31: Is `vscode.java.test.search.codelens` going to be removed  for god?

I use this in [nvim-jdtls](https://github.com/mfussenegger/nvim-jdtls) to provide `test_nearest_function` and `test_class`  methods

jdneo: Yes, the background is that VS Code is proposing a new set of testing APIs, so we are adopting to it. To match the data model, we need to update our backend logics.

But there is a new method to replace it, which will be added later. See: https://github.com/jdneo/vscode-java-test/blob/cs/new-test-api/java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestSearchUtils.java#L277-L325

mfussenegger: Thanks for the answer. I'm glad that there is going to be a replacement :+1:

jdneo: @mfussenegger 

Just FYI that the new refactored code has been merged into the main branch. The previous `vscode.java.test.search.codelens` is replaced by [findTestTypesAndMethods()](https://github.com/microsoft/vscode-java-test/blob/main/java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestSearchUtils.java#L291), actually their implementation logic is very similar.

PR #1239, opened 2021-07-28
Add load project logic at client side

jdneo:


src/controller/utils.ts
```diff
@@ -0,0 +1,74 @@
...
+        parent.children.add(item);
+    }
+    return item;
+}
+
+/**
+ * Parse the range object with server mode to client format
+ * @param range
+ */
+export function asRange(range: any): Range | undefined {
```
Eskibear, line 64: is it lsp2code?

jdneo: yes

PR #1240, opened 2021-07-28
List packages and classes at server side

jdneo:


java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/ProjectTestUtils.java
```diff
@@ -64,14 +64,39 @@
...
+    public static List<IClasspathEntry> getTestEntries(IJavaProject project) throws JavaModelException {
+        // Ignore default project
+        if (ProjectsManager.DEFAULT_PROJECT_NAME.equals(project.getProject().getName())) {
+            return Collections.emptyList();
+        }
+
+        final List<IClasspathEntry> entries = new LinkedList<>();
         for (final IClasspathEntry entry : project.getRawClasspath()) {
-            // Ignore default project
-            if (ProjectsManager.DEFAULT_PROJECT_NAME.equals(project.getProject().getName())) {
```
Eskibear, line 71: 😆

java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestSearchUtils.java
```diff
@@ -70,6 +86,102 @@
         return resultList;
     }
 
+    /**
+     * Return a list of test packages and types in the given project. The returned list has the following hierarchy:
```
Eskibear, line 90: so it's actually a pre-order traversal of the tree?

jdneo: Hmm, what I want to express here is that the elements in returned list might have children

Eskibear (approved):

PR #1244, opened 2021-07-29
Watch the workspace changes at client

jdneo:


java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestSearchUtils.java
```diff
@@ -382,7 +382,38 @@ private static void findTestItemsInTypeBinding(ITypeBinding typeBinding, JavaTes
...
+            }
+            final IJavaProject project = unit.getJavaProject();
+            if (project == null) {
+                return Collections.emptyList();
+            }
+            result.add(TestItemUtils.constructJavaTestItem(project, TestLevel.PROJECT, TestKind.None));
+
+            final IPackageFragment packageFragment = (IPackageFragment) unit.getParent();
+            if (packageFragment == null || !(packageFragment instanceof IPackageFragment)) {
+                return Collections.emptyList();
```
Eskibear, line 411: what about the project level item you just added into `result` above?

src/controller/testController.ts
```diff
@@ -21,6 +22,8 @@ export function createTestController(): void {
     testController.resolveHandler = async (item: TestItem) => {
         await loadChildren(item);
     };
+
+    startWatchingWorkspace();
```
Eskibear, line 26: what if I add folder into workspace?

java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestSearchUtils.java
jdneo: Just forget about it since only one project node is not enough to resolve the entire path from root to leaf

src/controller/testController.ts
jdneo: Good catch, I filed an issue to track this item. Would like to fix it systematically after all the components code are ready.

PR #1247, opened 2021-08-02
Resolve launch parameters for test runner

jdneo:


java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/launchers/JUnitLaunchConfigurationDelegate.java
```diff
@@ -167,30 +108,97 @@ private ITestKind getTestRunnerKind(ILaunchConfiguration configuration) {
...
+                        } else {
+                            parameters.add(paramTypeBinding.getQualifiedName());
+                        }
                     }
-                } catch (IOException | CoreException e) {
-                    // do nothing
                 }
-                return;
+                if (parameters.size() > 0) {
+                    testName += String.format("(%s)", String.join(",", parameters));
```
Eskibear, line 180: `"(" + String.join(",", parameters)) + ")" ` can be much faster.

Eskibear (approved):

java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/launchers/JUnitLaunchConfigurationDelegate.java
jdneo: Make sense!

PR #1249, opened 2021-08-02
Support find test location from full name

jdneo:


java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestSearchUtils.java
```diff
@@ -416,6 +425,131 @@ private static void findTestItemsInTypeBinding(ITypeBinding typeBinding, JavaTes
...
+    }
+
+    protected static final IType findType(final IJavaProject project, String className, IProgressMonitor monitor) {
+        final IType[] result = { null };
+        final String dottedName = className.replace('$', '.'); // for nested classes...
+        try {
+            if (project != null) {
+                result[0] = internalFindType(project, dottedName, new HashSet<IJavaProject>(), monitor);
+            }
+            if (result[0] == null) {
```
Eskibear, line 491: this branch is never visisted, as `project` is not null when you call `findType`.

jdneo: it's because `internalFindType()` might return null. In that case, use the search engine as a fallback

Eskibear: got it

Eskibear (approved):

PR #1255, opened 2021-08-03
Add changelog

jdneo:


package.json
```diff
@@ -262,7 +263,10 @@
         "test": "npm run compile && node ./dist/test/index.js",
         "lint": "gulp lint",
         "build-plugin": "gulp build-plugin",
-        "vscode:prepublish": "webpack --mode production"
+        "vscode:prepublish": "webpack --mode production",
+        "download-api": "vscode-dts dev -f",
```
jdneo, line 267: Temporarily need them since the API definition of `1.59.0` has not yet published to NPM registry

package.json
```diff
@@ -1,11 +1,12 @@
 {
+    "enableProposedApi": true,
```
Eskibear, line 2: remember to remove this when VS Code release latest stable version

Eskibear (approved):

package.json
jdneo: ok

Issue #1260, opened 2021-08-04
Error message is not clear when run TestNG project


chenyuanmac:
**Environment:**

OS: Linux
JDK: 11
Java Test Runner Extension: 0.31.0

**Repro steps:**

1. Download project:[http://www.mkyong.com/wp-content/uploads/2014/01/TestNG-Spring-Example.zip](http://www.mkyong.com/wp-content/uploads/2014/01/TestNG-Spring-Example.zip)
2. Open TestNG-Spring-Example project with VSCode
3. Click Run All Tests

**Result:**
Test cases cannot run
![image](https://user-images.githubusercontent.com/70991161/128156532-d71d4bff-e163-416a-b039-0b04b519b7f9.png)
Some error message show in test output.
![image](https://user-images.githubusercontent.com/70991161/128157058-765dc9d4-143e-4b60-9927-8e7c8eae1cd7.png)

jdneo:
To verify this issue:

1. Open a TestNG project, i.e. http://www.mkyong.com/wp-content/uploads/2014/01/TestNG-Spring-Example.zip
2. `Run All` tests
3. Check if the failure root cause is printed into the debug console

Eskibear:
verified that exceptions and callstack printed in debug console

PR #1285, opened 2021-08-18
Fixed the stack trace that is printed out when a TestNG test fails. This commit resolves issue #1260.

Kropie:


jdneo:
@Kropie Is there anything I can help for this work?

Issue #1283, opened 2021-08-16
Some tests inside static class won't run after click "Run All Tests" button


chenyuanmac:
**Environment:**

OS: Windows
JDK: 11
Java Test Runner Extension: 0.31.2

**Repro steps:**

1.git clone : [https://github.com/ZhengJun003/junit5-samples/tree/testrunner](https://github.com/ZhengJun003/junit5-samples/tree/testrunner)
2.Open junit5-jupiter-starter-maven with VSCode
3.Click "Run All Tests" from Test Explorer

**Result:**
Tests inside static class don't run.
![image](https://user-images.githubusercontent.com/70991161/129521199-cc4e08b4-6aef-4b99-874c-a9bdbdc22d8a.png)

Preranamisra:
Hey, can u add me as a contributor? I am working on it. Will solve this within somedays.

jdneo:
@Preranamisra You can just raise your PR directly if you would like to, and I'll take a look then. This repo is open for any kind of contribution.

jdneo:
@Preranamisra Is there anything I can help?

jdneo:
@cyberskycat We adopted the new testing APIs from VS Code. It's using the left side 'testing decorations' to run tests now

cyberskycat:
> @cyberskycat We adopted the new testing APIs from VS Code. It's using the left side 'testing decorations' to run tests now

ok , get it

Issue #1266, opened 2021-08-06
Put the test message location to where it happens


jdneo:
More specifically, if an assert fails, have the TestMessage's location of the assert match the location in the method for the given assert. Similarly if something triggers an exception, we could have a TestMessage for the exception in potentially even a different file - such as extract from the exception stack the location in testMethod() that resulted in the failure.

_Originally posted by @TimWebb in https://github.com/microsoft/vscode-java-test/issues/1162#issuecomment-893464998_

TimWebb:
@jdneo Your last two points make good sense, focusing on the assertion stack placement in the data model.

jdneo:
Verification steps:

1. Run a test case that will fail.
2. Check the test message displayed at where the error happens instead of the method declaration.

PR #1286, opened 2021-08-18
Fix stack trace output for TestNG failures

### LINKED ISSUE #1260: Error message is not clear when run TestNG project

Kropie:
This will fixes #1260.


jdneo:
Thank you for your contribution, I'll take a look this week

jdneo:
I tested the PR locally, but seems it's still not showing the errors according to the steps in #1260? Did I miss anything?

BTW, the refactoring looks good.

src/runners/testngRunner/TestNGRunnerResultAnalyzer.ts
```diff
@@ -40,7 +41,7 @@ export class TestNGRunnerResultAnalyzer implements IRunnerResultAnalyzer {
     }
 
     public analyzeData(data: string): void {
-        const lines: string[] = data.split(/\r?\n/);
+        const lines: string[] = this.unescape(data).split(/\r?\n/);
```
jdneo, line 44: I found the code in `analyzeData()` has some bugs, which I changed it to:

```js
public analyzeData(data: string): void {
    let match: RegExpExecArray | null;
    // tslint:disable-next-line: no-conditional-assignment
    while ((match = this.regex.exec(data)) !== null) {
        try {
           this.processData(match[1]);
        } catch (error) {
            this.testContext.testRun.appendOutput(`[ERROR] Failed to parse output data: ${match[1]}\n`);
        }
    }
}
```

Meanwhile, add the `g` flag to the regexp @ line 16:

```js
private readonly regex: RegExp = /@@<TestRunner-({[\s\S]*?})-TestRunner>/g;
```

jdneo:
@Kropie Would you like to update your PR?

Kropie:
@jdneo I will update my pull request tomorrow morning with the recommendations that you provided above. Thanks for your help on this!

src/runners/testngRunner/TestNGRunnerResultAnalyzer.ts
jdneo: Ah, I see why it's not working. Because here `.unescape(data)` is called before splitting to lines. Is there any reason to do this?

Kropie:
@jdneo  I've updated this PR to address the comments that you have above. Additionally I cleaned up the failure processing logic to clean up the "peek" view for test failures.

Kropie:
This is the result of the latest commit when running a failing test.
image

jdneo:
@Kropie The PR looks good now. the common logic is shared with this refactoring. But seems that it does not fix #1260, did I miss anything?

#1260 is a bug that when the suite is failed to run, nothing shows in UI except for the output channel. You can try download the project attached in #1260, and click run all.

Kropie:
@jdneo I may have missed a portion of the problem statement for #1260, but this does solve the formatting issues that were previously present when you would print out the results from the "peek error" screen. Additionally this resolves the stack trace issues in the preview.

Should the stack trace also be printing out on the debug console?

Kropie:
After looking running with the code linked in the pull request, this PR will resolve the formatting issues. That being said, if an exception is thrown while the tests are running all subsequent tests will not be able to run.

Kropie:
This is what I'm seeing when I run all the tests in that sample project
image

Kropie:
Finally figured out what I believe is the root of the problem. The tests in the example project use groups, but the test runner currently does not support selenium groups. I just pushed another commit that logs the error clearly to the user.

I think with the additional logging we have addressed the original issue in #1260 and an additional ticket should be written for adding TestNG group support to the java test runner extension.

Now when you run the tests you will see the following error message in the debug console.

Kropie:
This is an example of an error that is logged to the console to make things clearer to the user
image

jdneo:
The screenshot looks good. I'll take a deep look tomorrow.

> I think with the additional logging we have addressed the original issue in #1260 and an additional ticket should be written for adding TestNG group support to the java test runner extension.

Yes as you said, the current implementation of the TestNG runner has a lot of limitations. I'm hoping to migrate to the TestNG's official runner: https://github.com/testng-team/testng-remote, but so far does not have time to do this. 😥

src/runners/testngRunner/TestNGRunnerResultAnalyzer.ts
```diff
@@ -83,35 +79,15 @@ export class TestNGRunnerResultAnalyzer implements IRunnerResultAnalyzer {
...
-                        markdownTrace.appendText(line.replace(/</g, '&lt;').replace(/>/g, '&gt;'));
-                    }
-                    markdownTrace.appendText('\n');
-                }
                 const testMessage: TestMessage = new TestMessage(markdownTrace);
-                if (item.uri && item.range) {
-                    testMessage.location = new Location(item.uri, item.range);
+
+                for (const line of outputData.attributes.trace.split(/\r?\n/)) {
+                    this.processStackTrace(line, markdownTrace, testMessage, this.currentItem, this.projectName);
```
jdneo, line 89: Here `testMessage` is passed as `assertionFailure` but actually it's not.

`assertionFailure` is the test message looks like this:
image

you can see it's not a stack trace actually.

I guess that's why the stack trace now only have just few lines.

Kropie:
@jdneo Other than merging in the latest changes in from master is there anything else that needs to be done to merge and close this PR?

jdneo:
As I commented above, the refactor introduced a regression for the stacktrace of the TestNG cases. See: https://github.com/microsoft/vscode-java-test/pull/1286#discussion_r693457884

That needs to be resolved before merging.

Kropie:
I’ll resolve that tomorrow, thanks!

Kropie:
> As I commented above, the refactor introduced a regression for the stacktrace of the TestNG cases. See: [#1286 (comment)](https://github.com/microsoft/vscode-java-test/pull/1286#discussion_r693457884)
> 
> That needs to be resolved before merging.

Okay, @jdneo I believe I've resolved the issue with the stack trace that is being presented in the error message. I'm not positive what exactly it is supposed to look like but I've come up with what I believe is the right information.
### **JUnit**
image
### **TestNG**
image

src/runners/testngRunner/TestNGRunnerResultAnalyzer.ts
Kropie: Okay, @jdneo I believe I've resolved the issue with the stack trace that is being presented in the error message. I'm not positive what exactly it is supposed to look like but I've come up with what I believe is the right information.

jdneo:
I just tried with the new commit but I'm afraid it introduced new regression.

Previously, For JUnit, we can show the assertion failure via the diff view:
WeChat Screenshot_20210825094934

But now, it becomes to a plain text with a single line stack trace:
WeChat Screenshot_20210825094856

I guess you thought it a little bit complicated, base on this [commit](https://github.com/microsoft/vscode-java-test/pull/1286/commits/30565fe486b83b9f3a17f0b7221a1ed80353c79c), you can simply make the following change to make it work:

In TestNGRunnerResultAnalyzer, in the `else if (outputData.name === TEST_FAIL)` block :

```ts
...
if (outputData.attributes.trace) {
    ...
    for (const line of outputData.attributes.trace.split(/\r?\n/)) {
        this.processStackTrace(line, markdownTrace, undefined, this.currentItem, this.projectName);
    }

    const testMessage: TestMessage = new TestMessage(markdownTrace);
    if (item.uri && item.range) {
        testMessage.location = new Location(item.uri, item.range);
    }
    testMessages.push(testMessage);
}
const duration: number = Number.parseInt(outputData.attributes.duration, 10);
setTestState(this.testContext.testRun, item, this.currentTestState, testMessages, duration);
...
```

Just to provide more background information. We have two different test message:
1. The stack trace, which is the one you are dealing with
2. The assertion failure(expected ... but actual ...) showing with a diff view. This is the third argument of your extracted function `processStackTrace`. This is optional, in TestNG, we didn't support it, so you can just pass `undefined` to it.

> meanwhile, it's great that you also extracted the code to set the location of the test messages, but that can be addressed in another PR (to avoid this one become bigger and bigger endlessly).

Kropie:
> I just tried with the new commit but I'm afraid it introduced new regression.
> 
> Previously, For JUnit, we can show the assertion failure via the diff view:
> WeChat Screenshot_20210825094934
> 
> But now, it becomes to a plain text with a single line stack trace:
> WeChat Screenshot_20210825094856
> 
> I guess you thought it a little bit complicated, base on this [commit](https://github.com/microsoft/vscode-java-test/pull/1286/commits/30565fe486b83b9f3a17f0b7221a1ed80353c79c), you can simply make the following change to make it work:
> 
> In TestNGRunnerResultAnalyzer, in the `else if (outputData.name === TEST_FAIL)` block :
> 
> ```ts
> ...
> if (outputData.attributes.trace) {
>     ...
>     for (const line of outputData.attributes.trace.split(/\r?\n/)) {
>         this.processStackTrace(line, markdownTrace, undefined, this.currentItem, this.projectName);
>     }
> 
>     const testMessage: TestMessage = new TestMessage(markdownTrace);
>     if (item.uri && item.range) {
>         testMessage.location = new Location(item.uri, item.range);
>     }
>     testMessages.push(testMessage);
> }
> const duration: number = Number.parseInt(outputData.attributes.duration, 10);
> setTestState(this.testContext.testRun, item, this.currentTestState, testMessages, duration);
> ...
> ```
> 
> Just to provide more background information. We have two different test message:
> 
> 1. The stack trace, which is the one you are dealing with
> 2. The assertion failure(expected ... but actual ...) showing with a diff view. This is the third argument of your extracted function `processStackTrace`. This is optional, in TestNG, we didn't support it, so you can just pass `undefined` to it.
> 
> > meanwhile, it's great that you also extracted the code to set the location of the test messages, but that can be addressed in another PR (to avoid this one become bigger and bigger endlessly).

First, thanks for your patience!
Second, I was able to resolve the regression with the "Expected ... but actual ... logic" that you referenced above.

I pushed the update out based on my last commit because the delta was rather small. Let me know if you would still rather have me do the code update that you referenced above on the previous commit.

jdneo:
Thanks for the quite response! Personally, I would prefer to reset to the commit I mentioned and append commit based on that. (To make the total change as simple as possible). We can definitely address other issues by several PRs separately.

Kropie:
> Thanks for the quite response! Personally, I would prefer to reset to the commit I mentioned and append commit based on that. (To make the total change as simple as possible). We can definitely address other issues by several PRs separately.

Just tried to make the modification to the [commit](https://github.com/microsoft/vscode-java-test/pull/1286/commits/30565fe486b83b9f3a17f0b7221a1ed80353c79c) referenced above, but that has issues that are addressed by my latest commit. Specifically if you have a stack trace with multiple elements in it and you have a failure message instead of a "expected .. but actual ..." message you will only see the first element on the stack at the time of the failure.
<br>
Additionally for TestNG tests you only ever get the first element on the stack at the time of failure.
<br>
Also, for help visualizing the problem:
### Limited Stack Trace (Update of referenced [commit](https://github.com/microsoft/vscode-java-test/pull/1286/commits/30565fe486b83b9f3a17f0b7221a1ed80353c79c))
image

### Full Stack Trace (Current latest commit in this PR)
image

Kropie:
That being said I am okay with updating that previous commit and just creating another PR if that is your preference.

Kropie:
That being said I am okay with updating that previous commit and just creating another PR if that is your preference.

jdneo:
I prefer to resolve them separately in differently. So this PR is aiming to resolve #1283 & #1291.

I believe the problem you mentioned is similar to #1266, and we can for sure resolve that problem in another PR.

Please correct me if I misunderstand

Kropie:
That works for me. I'll have the pull request updated tomorrow

jdneo (approved):

jdneo:
@Kropie, looks pretty good! Thank you for the nice job!

Kropie:
> @Kropie, looks pretty good! Thank you for the nice job!

Thanks for the guidance!

PR #1295, opened 2021-08-23
fix: Cannot see full stack trace in TestNG tests

### LINKED ISSUE #1291: Cannot see full stacktrace in TestNG tests
Hi, this is my env
image
and I'm using Test Runner for Java v0.31.2

I'm working on a Spring project and while I'm having Failed to load ApplicationContext Error I can't see the reason.
image
 It seems like the stack trace is partial.

Also, I can't see the stack trace in the Debug Console because I'm getting a lot of logs while an error occurs and the log has a limited number of lines, so I can't see the relevant error because it was removed.

In the last version I could see the full stack trace in the report, and to tell the truth, it was easier to look at the report because it was on a separate page.

jdneo:
fix #1291


jdneo:
Already fixed in #1286

Issue #878, opened 2019-11-18
Failed to get workspace folder for the test items


JdKock:
I have created a default spring-boot project to see how it works in VSCode for Java development.
I started VSCode and via File - Open Folder, opened my new Java spring-boot project.
The project runs without any problems and when I opened a test in the editor I see the "Run Test/Debug Test" shortcuts above the test class.
When I now click on "Run Test" I get the error message: "**Failed to get workspace folder for the test items**"

I also don't see any tests in the Test explorer. Even after adding the following to my pom file.
<testSourceDirectory>${project.basedir}/src/test/java/</testSourceDirectory>

I really like to try to use VSCode for my Java development but it looks like it not yet ready for this. There are a lot of issues I see here about testing.

Can anyone help me to get my tests working ?

jdneo:
Hi @JdKock，

Is there any repro steps that I can try in my side?

jdneo:
Hi @JdKock 

Is the problem still occur?

JdKock:
No it's working now. Thnx.

AndrzejPajda:
@JdKock 
Can you share how did you solve this issue? My CodeLens is properly recognizing test classes but not showing them in the test outliner. I am also getting 'Failed to get workspace folder for test items.' error when trying to run any tests.

adchang:
@AndrzejPajda 
Any progress on resolving this? I have the same exact issue. I opened my Maven project in VS Code (working in Eclipse). CodeLens shows Run/Debug Test, but Test Explorer is empty. When I click Run or Debug Test, I am getting the "Failed to get workspace folder for test items"

Folder is added to workspace; .classpath is correct (with the attribute test); tried testSourceDirectory in pom; java.autobuild.enabled = auto; ran Update project configuration; cleaned java server workspace; running mvn test works

In Java Test Runner, I'm getting the error "Test job is canceled.Failed to clean up. TypeError: this.testIds is not iterable"...not much info googling this error

jdneo:
@AndrzejPajda @adchang

Are you just running your project in your local environment? Or in a `Remote` environment?

Would you mind the share the project to us for diagnostic?

adchang:
This is happening to me in a local environment and a new mvn project

Thanks for looking into it!

********************************************************************************
Version: 1.44.2
Commit: ff915844119ce9485abfe8aa9076ec76b5300ddd
Date: 2020-04-16T17:50:03.709Z
Electron: 7.1.11
[...]

PR #1317, opened 2021-10-11
#878 Fix some 'Failed to get workspace folder' error conditions.

rbi:
This fixes my instance of the #878 error. Also the bug is closed it still ocures in my environment. I've described my problem there in comment https://github.com/microsoft/vscode-java-test/issues/878#issuecomment-938396359.

As the bug is already closed I'm not sure if someone will notice my comments there. I also did not want to open another bug.

### DIFF
--- a/src/controller/testController.ts
+++ b/src/controller/testController.ts
@@ -195,8 +195,10 @@ export const runTests: (request: TestRunRequest, option: IRunOption) => any = in
                             return resolve();
                         });
                         option.progressReporter?.report('Resolving launch configuration...');
+                        // This normalization fixes issues with upper case vs. lower case drive letters on Windows. See #878.
+                        const uri = items[0].uri?.fsPath? Uri.file(items[0].uri?.fsPath!) : items[0].uri!;
                         // TODO: improve the config experience
-                        const workspaceFolder: WorkspaceFolder | undefined = workspace.getWorkspaceFolder(items[0].uri!);
+                        const workspaceFolder: WorkspaceFolder | undefined = workspace.getWorkspaceFolder(uri);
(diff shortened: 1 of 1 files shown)


jdneo:
Thank you @rbi, will take a look

jdneo:
Actually I'm a little bit confused about this change.

When the URI passed to `workspace.getWorkspaceFolder()` is upper case for the device segment, then it's not working? Did I understand correctly?

rbi:
Yes that seams to be the case. The device segment of the `path` field of the URI must be upper-case for the bug to appear. When you do `toString` on such an uri it shows the drive segment lower-case although it is upper-case in the `path` field.

I've got the idea for the workaround from here:
https://github.com/microsoft/vscode/issues/36221#issuecomment-337670295

This bug should be fixed VS Code 1.18 so that the  `vscode.workspace.getWorkspaceFolder` method can handle upper-case drive segements. In my environment it still is an issue. I do remote development on a remote Windows. So maybe the `vscode.workspace.getWorkspaceFolder` is re-implemented somewhere in the remote development extensions where they implemented the bug a second time? If so it may should be fixed there.

Other specialties about my system:
* The VS-Code GUI is running on a Linux machine connecting to a Windows VM with a UI session remotely.
* I start the Windows sshd not via service but manually in a Powershell window as I want to be able to start GUI programs from VS-Code.

jdneo:
@rbi seems the upstream issue has been fixed. Is this PR still valid?

rbi:
I haven't tried the fix yet but as they could reproduce the problem I guess it should work now. At least it seams to be no problem in the vscode-java-test plugin as you said.

PR #1323, opened 2021-10-25
feat: Show stack trace to where error occurs

### LINKED ISSUE #1266: Put the test message location to where it happens

jdneo:
fix #1266 

If the stacktrace contains location belonging to the source file, then we show it at that location.

https://user-images.githubusercontent.com/6193897/138637985-c0773467-697e-41c9-b127-e242af17ca8e.mp4


Eskibear (approved):
LGTM

PR #1326, opened 2021-11-02
feat: Add 'envFile' in test configuration

### LINKED ISSUE #1214: Allow `envFile` alternative to `env` in `java.test.config`
The VS Code `launch.json` file allows specifying `envFile` in addition to `env`. However, there does not appear to be a similar mechanism when running tests via this extension. It would be super handy to be able to have 1 file specify the `env` rather than duplicate it in both `.env` and `.vscode/settings.json`.

jdneo:
resolve #1214

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -40,6 +40,9 @@ jobs:
     - name: Build OSGi bundle
       run: npm run build-plugin
 
+    - name: prepublish
+      run: npm run vscode:prepublish
+
     - name: Test extension
       run: DISPLAY=:99 npm test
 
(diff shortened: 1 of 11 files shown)


.github/workflows/build.yml
```diff
@@ -40,6 +40,9 @@ jobs:
     - name: Build OSGi bundle
       run: npm run build-plugin
 
+    - name: prepublish
+      run: npm run vscode:prepublish
```
CsCherrYY, line 44: Suggest moving this part to script `npm run test` in package.json, to make it more convenient in development.

jdneo: Make sense

CsCherrYY: get it

CsCherrYY (dismissed):
please do not forget to update the wiki

CsCherrYY (approved):

PR #1328, opened 2021-11-03
feat: Support customize the classpath in test configuration

### LINKED ISSUE #1140: Unable to define a custom classpath for a Junit test
Hi,
Some tests require custom classpath to run, when having a workspace with multiple projects (defined with .project and .classpath like in eclipse) and dependency inversion between some projects (plugin architecture).

Like in eclipse we need to define a launch configuration with a custom classpath made of path or project name.
We didn't found a way to do it in [Run with Configuration](https://github.com/Microsoft/vscode-java-test/wiki/Run-with-Configuration).

Using "-cp" in vmArgs or "CLASSPATH" env variable didn't work.

jdneo:
resolve #1140 


package.json
```diff
@@ -176,6 +176,30 @@
                             "description": "%configuration.java.test.config.workingDirectory.description%",
                             "default": "${workspaceFolder}"
                         },
+                        "classPaths": {
```
testforstephen, line 179: should add `modulePaths` as well.

jdneo: Yes, there will be another PR to solve the module path  issue: #1101

testforstephen: i assume you just pass these configs to debugger directly. Is there anything to block you adding modulePaths at the moment?

jdneo: No, just want to have two PRs because they are two different issues. But I'm also fine to address them together in this single PR.

testforstephen (approved):

PR #1349, opened 2021-12-23
chore: Migrate to Java 11

jdneo:
Now that the Java Language Support Extension has embedded JRE 17 to run all Java extensions, it's time to migrate to Java 11.

### DIFF
--- a/java-extension/com.microsoft.java.test.plugin/META-INF/MANIFEST.MF
+++ b/java-extension/com.microsoft.java.test.plugin/META-INF/MANIFEST.MF
@@ -4,7 +4,7 @@ Bundle-Name: com.microsoft.java.test.plugin
 Bundle-SymbolicName: com.microsoft.java.test.plugin;singleton:=true
 Bundle-Version: 0.33.1
 Bundle-Activator: com.microsoft.java.test.plugin.util.JUnitPlugin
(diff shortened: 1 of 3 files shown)


testforstephen:
The compiler version in some pom.xml is still 1.8.

https://github.com/microsoft/vscode-java-test/blob/78717ac77a2168372d6109d6fad82ad439e93900/java-extension/com.microsoft.java.test.runner/pom.xml#L30-L35

jdneo:
> The compiler version in some pom.xml is still 1.8.

That's the test runner for TestNG, which is just a common Java Application, not a OSGi bundle. So I think it's fine to keep it as `1.8`

testforstephen:
I don't see you specify compiler version explicitly in your plugin pom.xml, I'm curious what's the target Java version when you're compile artifacts from cmd.

If there are 1.8 and 11 mixed in the final build artifacts, I'd rather not update to 11.

karlvr:
@jdneo This line in `JUnitLaunchConfigurationDelegate.java` uses `Set.of` which is not available in Java 8, but the maven compiler plugin is set to use Java 1.8:

```java
private static final Set<String> testNameArgs = Set.of("-test", "-classNames", "-packageNameFile", "-testNameFile");
```

So I think the Java version in the `pom.xml` referenced above might need to be updated to 11?

Issue #1367, opened 2022-01-25
Cannot run test class with '$' in its name sometime.


jdneo:
Looks like the upstream test runner could not deal with this case decently.

I can see traces:

```
org.eclipse.core.runtime.CoreException: No tests found with test runner 'JUnit 4'.
	at org.eclipse.jdt.junit.launcher.JUnitLaunchConfigurationDelegate.abort(JUnitLaunchConfigurationDelegate.java:763)
	at org.eclipse.jdt.junit.launcher.JUnitLaunchConfigurationDelegate.evaluateTests(JUnitLaunchConfigurationDelegate.java:372)
	at org.eclipse.jdt.junit.launcher.JUnitLaunchConfigurationDelegate.getVMRunnerConfiguration(JUnitLaunchConfigurationDelegate.java:162)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
	at java.base/java.lang.reflect.Method.invoke(Unknown Source)
	at com.microsoft.java.test.plugin.launchers.JUnitLaunchConfigurationDelegate.getJUnitLaunchArguments(JUnitLaunchConfigurationDelegate.java:75)
	at com.microsoft.java.test.plugin.launchers.JUnitLaunchUtils.resolveLaunchArgument(JUnitLaunchUtils.java:102)
	at com.microsoft.java.test.plugin.handler.TestDelegateCommandHandler.executeCommand(TestDelegateCommandHandler.java:50)
	at org.eclipse.jdt.ls.core.internal.handlers.WorkspaceExecuteCommandHandler$1.run(WorkspaceExecuteCommandHandler.java:215)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:45)
	at org.eclipse.jdt.ls.core.internal.handlers.WorkspaceExecuteCommandHandler.executeCommand(WorkspaceExecuteCommandHandler.java:205)
	at org.eclipse.jdt.ls.core.internal.handlers.JDTLanguageServer.lambda$4(JDTLanguageServer.java:507)
	at org.eclipse.jdt.ls.core.internal.BaseJDTLanguageServer.lambda$0(BaseJDTLanguageServer.java:75)
[...]

PR #1368, opened 2022-01-25
fix: Generate '_' instead of '$' for inner class

jdneo:
- The upstream runner could not fine tests when '$' is in class name,
  so we use '_' when generating test classes.

// see discussion in #1367

### DIFF
--- a/java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestGenerationUtils.java
+++ b/java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestGenerationUtils.java
@@ -408,8 +408,8 @@ public class TestGenerationUtils {
 
     private static String getDefaultTestFullyQualifiedName(ITypeBinding typeBinding, IJavaProject project,
             IClasspathEntry testEntry) throws JavaModelException {
-        final String defaultName = typeBinding.getBinaryName() + "Test";
-        final String attemptName = typeBinding.getBinaryName() + "Tests";
+        final String defaultName = getClassName(typeBinding) + "Test";
+        final String attemptName = getClassName(typeBinding) + "Tests";
         try {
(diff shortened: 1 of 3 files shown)


java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestGenerationUtils.java
```diff
@@ -444,6 +444,17 @@ private static String getDefaultTestFullyQualifiedName(ITypeBinding typeBinding,
         return defaultName;
     }
 
+    private static String getClassName(ITypeBinding typeBinding) {
```
testforstephen, line 447: How about simply using `typeBinding.getBinaryName().replaceAll("\\$", "_")`?

jdneo: I'm afraid that will not only replace '$' as a splitter, but also '$' as part of the name. (Though that should not be a very common case).

testforstephen: > The upstream runner could not fine tests when '$' is in class name,

If the class itself has '$' in its name, does the runner support it?

jdneo: Oh I got your point. Yes if upstream cannot handle this case well, there is no point to keep `$` in the class name.

jdneo: @testforstephen PR updated. new code replaces all the `$` in type name to `_`, but reserve the `$` in packages, since the `$` in packages won't affect the execution.

testforstephen (approved):

PR #1382, opened 2022-03-03
fix: Make sure extension could be activated

### LINKED ISSUE #1381: Tests are not found anymore since 0.34.1
When using version 0.34.1 (I did so on Gitpod) none of my tests are recognized anymore.
No play button is shown and  also in the Testing View no tests are shown. Everything works fine with 0.34.0

jdneo:
fix #1381 

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -20,6 +20,7 @@ import { registerAskForChoiceCommand, registerAdvanceAskForChoice, registerAskFo
 import { enableTests } from './commands/testDependenciesCommands';
 
 export let extensionContext: ExtensionContext;
+let componentsRegistered: boolean = false;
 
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -69,13 +59,30 @@ async function doActivate(_operationId: string, context: ExtensionContext): Prom
             }));
         }
 
+        if (extensionApi.onDidClasspathUpdate) {
```
CsCherrYY, line 62: Why do we change the sequence of event listeners here?

jdneo: Just would like to register the server mode change event more earlier

CsCherrYY: get it

CsCherrYY (approved):

PR #1389, opened 2022-03-24
prepare for pre-release

Eskibear:
pipeline is configured in classic way. engine is already 1.65

### DIFF
--- a/scripts/prepare-nightly-build.js
+++ b/scripts/prepare-nightly-build.js
@@ -1,0 +1,25 @@
+const fs = require("fs");
+
+const json = JSON.parse(fs.readFileSync("./package.json").toString());
+const stableVersion = json.version.match(/(\d+)\.(\d+)\.(\d+)/);
+const major = stableVersion[1];
+const minor = stableVersion[2];
+
+function prependZero(number) {
+    if (number > 99) {
+        throw "Unexpected value to prepend with zero";
+    }
+    return `${number < 10 ? "0" : ""}${number}`;
+}
+
+const date = new Date();
+const month = date.getMonth() + 1;
+const day = date.getDate();
+const hours = date.getHours();
+patch = `${date.getFullYear()}${prependZero(month)}${prependZero(day)}${prependZero(hours)}`;
+
+const insiderPackageJson = Object.assign(json, {
+    version: `${major}.${minor}.${patch}`,
+});
+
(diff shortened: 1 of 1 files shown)


jdneo:
Does that mean all the things will automatically work once this PR is merged?

CsCherrYY:
> Does that mean all the things will automatically work once this PR is merged?

we can check if the property `Microsoft.VisualStudio.Code.PreRelease` is set to true in the output bits to verify if it works

CsCherrYY (approved):

testforstephen:
this just bumps the package.json version, but not updating the Java bundle version. If I change Java plugin code, does it really work?

Eskibear:
> If I change Java plugin code, does it really work?

Good point. AFAIK current impl in jdtls won't reinstall the bundle.

Eskibear:
@jdneo @CsCherrYY  If it's the case, we should probably update the script, to also update version of *.jar and the reference in `javaExtensions`.

@testforstephen Does jdtls cache the bundle by name + version, or by absolute path? If it's the path, then full path of the jar is different.

testforstephen:
If the bundle version and symbolic are same, it won't install your new jar but reuse the old one.
https://github.com/eclipse/eclipse.jdt.ls/blob/4dd3f74d3883586bbf10d29b218633efe9bee7a2/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/handlers/BundleUtils.java#L126

One idea is to set the bundle version to 0.34.2.qualifier, the tycho will auto generate a timestamp on the jar name. Meanwhile we need a new script to update jar reference in package.json.

https://github.com/microsoft/vscode-java-test/blob/b45f1ca4f567c1c68d0bf4f641172b5a187a91c9/java-extension/com.microsoft.java.test.plugin/META-INF/MANIFEST.MF#L5-L9

https://github.com/microsoft/vscode-java-test/blob/b45f1ca4f567c1c68d0bf4f641172b5a187a91c9/package.json#L47-L52

CsCherrYY:
> Meanwhile we need a new script to update jar reference in package.json.

If we use `qualifier` we will have to update the jar name when publishing both the stable version and the pre-release version. This script will help them both.

testforstephen:
Re-look at the jdt.ls bundle loading mechanism again, it seems if the bundle location is different, it will uninstall the old bundle and install the new location again. This PR should be OK.

PR #1402, opened 2022-04-07
feat: Simplify the stacktrace

### LINKED ISSUE #1281: [Request] Hide the JUnit Callstack from the stacktrace
There are (few, but there are) cases in which I'd like to take a look at the callstack for a _failed_ (as opposed errored) test. Mainly when using assertThrows. However I would like to have it shortened by default so that in the simplest case every line starting with `at org.junit` is replaced with the `... X more lines` just like the end is shortened right now.

I'd be satisfied with a setting like "shorten lines matching this regex from the stack trace".

For example when assertThrows fails, because instead of ExpectedException I got DifferentException, I'd love the output to look like
```
org.opentest4j.AssertionFailedError: Unexpected exception type thrown ==> expected: <org.example.ExpectedException> but was: <org.example.DifferentException>
    at org.junit.jupiter.api.AssertThrows.assertThrows(AssertThrows.java:65)
    ... 70 more lines
Caused by: org.example.DifferentException: Message
    at x
    at y
    at z
```

jdneo:
- Simplify the stacktrace by excluding internal traces, which defined in
  getStacktraceFilter().

fix #1281

### DIFF
--- a/src/runners/baseRunner/RunnerResultAnalyzer.ts
+++ b/src/runners/baseRunner/RunnerResultAnalyzer.ts
@@ -31,8 +31,8 @@ export abstract class RunnerResultAnalyzer {
             }
 
             const location: string = traceResults[4];
-            let sourceName: string | undefined = undefined;
-            let lineNumLiteral: string | undefined = undefined;
+            let sourceName: string | undefined;
+            let lineNumLiteral: string | undefined;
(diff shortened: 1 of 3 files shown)


jdneo:
Before:
image

After:
image

Eskibear (approved):

PR #1403, opened 2022-04-07
fix: Always display the test message in the test's range

### LINKED ISSUE #1392: Error message won't append at the method
Sometimes the error message won't append at the method. For example, when the error happens at the `@before` lifecycle, it will add to the before method, which is wired.

jdneo:
fix #1392

The test message should always display inside the range of a test message. Even when its stacktrace locates outside of it. In this case, we simply set the test message location at the first line of the test case.


jdneo:
Before
before

It's hard to find the test message when the file is large or multiple cases failed at the same place.

After
after

Eskibear (approved):

PR #1407, opened 2022-04-17
feat: re-run single JUnit5 parameterized test invocation via uniqueId

fladdimir:
Support for re-running/debugging a parameterized JUnit5 test with a single set of parameters, similar to the corresponding eclipse feature:
vscode_rerun_possible

resolve https://github.com/microsoft/vscode-java-test/issues/1408

### DIFF
--- a/java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/launchers/JUnitLaunchConfigurationDelegate.java
+++ b/java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/launchers/JUnitLaunchConfigurationDelegate.java
@@ -17,6 +17,7 @@ import com.microsoft.java.test.plugin.model.TestLevel;
 import com.microsoft.java.test.plugin.util.JUnitPlugin;
 import com.microsoft.java.test.plugin.util.TestSearchUtils;
 
+import org.apache.commons.lang3.StringUtils;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
@@ -182,7 +183,7 @@ public class JUnitLaunchConfigurationDelegate extends org.eclipse.jdt.junit.laun
             }
             arguments.add(method.getDeclaringType().getFullyQualifiedName() + ':' + testName);
 
-            if (this.args.uniqueId != null && !this.args.uniqueId.isBlank()) {
+            if (StringUtils.isNotBlank(this.args.uniqueId)) {
                 arguments.add("-uniqueId");
                 arguments.add(this.args.uniqueId);
             }
--- a/java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/launchers/JUnitLaunchConfigurationDelegate.java
+++ b/java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/launchers/JUnitLaunchConfigurationDelegate.java
@@ -181,6 +181,11 @@ public class JUnitLaunchConfigurationDelegate extends org.eclipse.jdt.junit.laun
                 }
             }
             arguments.add(method.getDeclaringType().getFullyQualifiedName() + ':' + testName);
+
+            if (this.args.uniqueId != null && !this.args.uniqueId.isBlank()) {
+                arguments.add("-uniqueId");
+                arguments.add(this.args.uniqueId);
+            }
         }
     }
 
--- a/src/controller/testController.ts
+++ b/src/controller/testController.ts
@@ -302,12 +302,27 @@ function handleInvocationRerun(testItems: TestItem[]): boolean {
         }
     });
 
-    if (testItems.length === 1 && dataCache.get(testItems[0])?.testLevel === TestLevel.Invocation) {
+    // the test-explorer supports selecting multiple items,
+    // but when re-running an invocation only selecting a single item is supported
+    const numInvocations: number = testItems.filter((item: TestItem) => dataCache.get(item)?.testLevel === TestLevel.Invocation).length;
+    if (numInvocations > 0) {
+        if (numInvocations < testItems.length) {
+            window.showErrorMessage('When re-running an invocation of a parameterized test, no other tests must be selected. Please select either a single invocation or no invocations at all.');
+            testItems.length = 0;
+            return isInvocationRerun;
+        }
+        if (numInvocations > 1) {
+            window.showErrorMessage('Re-running multiple invocations of a parameterized test is not supported, please select only one invocation at a time.');
+            testItems.length = 0;
+            return isInvocationRerun;
+        }
         // if a single invocation is to be rerun,
         // we run the parent method, but with restriction to the single invocation parameter-set
         const invocation: TestItem = testItems[0];
         if (!invocation.parent || !dataCache.get(invocation.parent)) {
-            sendError(new Error('Trying to re-run a single test invocation, but could not find a corresponding method-level parent item with data.'));
+            const errMsg: string = 'Trying to re-run a single test invocation, but could not find a corresponding method-level parent item with data.';
+            sendError(new Error(errMsg));
+            window.showErrorMessage(errMsg);
             testItems.length = 0;
             return isInvocationRerun;
         }
--- a/src/controller/testController.ts
+++ b/src/controller/testController.ts
@@ -249,14 +249,14 @@ export const runTests: (request: TestRunRequest, option: IRunOption) => any = in
  * Set all the test item to queued state
  */
 function enqueueTestMethods(testItems: TestItem[], run: TestRun): void {
-   const queuedTests: TestItem[] = [...testItems];
-   while (queuedTests.length) {
-       const queuedTest: TestItem = queuedTests.shift()!;
-       run.enqueued(queuedTest);
-       queuedTest.children.forEach((child: TestItem) => {
-           queuedTests.push(child);
-       });
-   }
+    const queuedTests: TestItem[] = [...testItems];
+    while (queuedTests.length) {
+        const queuedTest: TestItem = queuedTests.shift()!;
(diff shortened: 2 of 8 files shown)


jdneo:
Hi @fladdimir, thank you for your contribution.

I saw the PR is closed. Did you meet any problem or need any help?

fladdimir:
> Hi @fladdimir, thank you for your contribution.
> 
> I saw the PR is closed. Did you meet any problem or need any help?

Hi @jdneo, creating it here was unintentional, I was not sure about the preferred workflow. I will update it and open it again.

jdneo:
Ok, if you want to contribute a change, directly create a PR is fine. :)

src/controller/testController.ts
```diff
@@ -265,23 +265,47 @@ function enqueueTestMethods(testItems: TestItem[], run: TestRun): void {
...
+        testItems.push(...request.include);
+    } else {
+        testController?.items.forEach((item: TestItem) => {
+            testItems.push(item);
+        });
+    }
+    if (testItems.length === 0) {
+        return [];
+    }
+    handleInvocationRerun(testItems);
```
fladdimir, line 279: (the only non-whitespace change in this method)

jdneo:
I will review this PR these days, thank you!

java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/launchers/JUnitLaunchConfigurationDelegate.java
```diff
@@ -181,6 +181,11 @@ private void addTestItemArgs(List<String> arguments) throws CoreException {
                 }
             }
             arguments.add(method.getDeclaringType().getFullyQualifiedName() + ':' + testName);
+
+            if (this.args.uniqueId != null && !this.args.uniqueId.isBlank()) {
```
jdneo, line 185: We can use `StringUtils.isNotBlank()` here

src/controller/testController.ts
```diff
@@ -265,23 +265,47 @@ function enqueueTestMethods(testItems: TestItem[], run: TestRun): void {
...
+            dataCache.get(item)!.uniqueId = undefined;
+        }
+    });
+
+    if (testItems.length === 1 && dataCache.get(testItems[0])?.testLevel === TestLevel.Invocation) {
+        // if a single invocation is to be rerun,
+        // we run the parent method, but with restriction to the single invocation parameter-set
+        const invocation: TestItem = testItems[0];
+        if (!invocation.parent || !dataCache.get(invocation.parent)) {
+            sendError(new Error('Trying to re-run a single test invocation, but could not find a corresponding Method-level parent item with data.'));
```
jdneo, line 302: nit: Method-level -> method-level

src/controller/testController.ts
```diff
@@ -265,23 +265,47 @@ function enqueueTestMethods(testItems: TestItem[], run: TestRun): void {
...
+    const excludingItems: TestItem[] = await expandTests(request.exclude || [], TestLevel.Class, token);
+    testItems = _.differenceBy(testItems, excludingItems, 'id');
+    return testItems;
+}
+
+function handleInvocationRerun(testItems: TestItem[]): void {
+
+    // always remove uniqueIds from non-invocation items, since they would have been set for a past run
+    testItems.forEach((item: TestItem) => {
+        if (dataCache.get(item) && dataCache.get(item)!.testLevel !== TestLevel.Invocation) {
```
jdneo, line 292: How about extract `dataCache.get(item)` to a local variable? That would make the code here more neat.

src/runners/junitRunner/JUnitRunnerResultAnalyzer.ts
```diff
@@ -226,6 +226,9 @@ export class JUnitRunnerResultAnalyzer extends RunnerResultAnalyzer {
             const parentIndex: string = result[5];
             const displayName: string = result[6].replace(/\\,/g, ',');
 
+            const uniqueId: string | undefined = this.testContext.kind === TestKind.JUnit5 ?
```
jdneo, line 229: Let remove the empty line at 228, since uniqueId should be grouped with the above set of code logically.

jdneo:
One more comment about UX, I noticed that when running a parameterized invocation, other invocations will be removed from parent node, can we avoid that?

Maybe it's because all invocations are removed here: https://github.com/microsoft/vscode-java-test/blob/7b8b5b02d62b20bb25707e35b5e1acdfd00d30c6/src/controller/testController.ts#L279

a

fladdimir:
> One more comment about UX, I noticed that when running a parameterized invocation, other invocations will be removed from parent node, can we avoid that?

Good hint, now the invocations are not removed anymore in case of a single re-run. During result processing, the invocation item is then updated instead of created.

retain_test_invocations_on_rerun

src/controller/testController.ts
```diff
@@ -265,23 +265,57 @@ function enqueueTestMethods(testItems: TestItem[], run: TestRun): void {
...
+        if (itemData && itemData.testLevel !== TestLevel.Invocation) {
+            itemData.uniqueId = undefined;
+        }
+    });
+
+    if (testItems.length === 1 && dataCache.get(testItems[0])?.testLevel === TestLevel.Invocation) {
+        // if a single invocation is to be rerun,
+        // we run the parent method, but with restriction to the single invocation parameter-set
+        const invocation: TestItem = testItems[0];
+        if (!invocation.parent || !dataCache.get(invocation.parent)) {
```
jdneo, line 309: Besides `sendError()`, maybe we can also show an error message to let users know there is an error?

```ts
const errMsg = 'Trying to re-run a single test invocation, but could not find a corresponding method-level parent item with data.';
sendError(new Error(errMsg));
window.showErrorMessage(errMsg);
```

src/controller/testController.ts
```diff
@@ -265,23 +265,57 @@ function enqueueTestMethods(testItems: TestItem[], run: TestRun): void {
...
+    testItems = _.differenceBy(testItems, excludingItems, 'id');
+    return testItems;
+}
+
+/**
+ * Check and preparation in case a single invocation of a parameterized test shall be re-run.
+ * @param testItems
+ * @returns true if a single invocation shall be re-run, otherwise false
+ */
+function handleInvocationRerun(testItems: TestItem[]): boolean {
```
jdneo, line 294: There is another issue might happen here. The current Testing explorer supports multi-selection. Users may select two invocation nodes and run them, which is not supported on our side now. Can we add one more check here? If there are more than 1 invocation nodes selected, we skip the test run and pop-up a message telling user that it's not supported?

jdneo (commented):
Thank you @fladdimir. It looks pretty well! Just a very few small things to be addressed. After that I think it's good to merge.

src/controller/testController.ts
fladdimir: I added the check and an additional one to handle the case of selecting a mix of invocation and non-invocation items. The behavior there could be improved (e.g. checking whether the invocation is already included in the selected parts of the test-tree). However this may be fine for supporting the primary use-case of this feature?

jdneo:
Some new findings:

For this sample project: https://github.com/junit-team/junit5-samples/blob/main/junit5-jupiter-starter-maven/src/test/java/com/example/project/CalculatorTests.java

When I change the parameter value and rerun the invocation, the display name won't be updated.

a

The root cause is that here: https://github.com/microsoft/vscode-java-test/blob/a66e066b6ea70904ef933d7fb76d41a58eb8a98d/src/runners/junitRunner/JUnitRunnerResultAnalyzer.ts#L238.

We can keep the original implementation which always creating a new test item instead of updating an existing one. By reverting the change, the displayname will be updated.

Another issue I found is after running a parameterized tests, return it from the method level will throw an error.
This is because VS Code will give us the method item and all its invocation nodes. Seems that we need to try to merge the test items before the statement `const numInvocations: number = testItems.filter((item: TestItem) => dataCache.get(item)?.testLevel === TestLevel.Invocation).length;`.

fladdimir:
> When I change the parameter value and rerun the invocation, the display name won't be updated.
This is should be fixed now by also updating the item.label.

src/controller/utils.ts
```diff
@@ -117,6 +117,7 @@ function updateOrCreateTestItem(parent: TestItem, childData: IJavaTestItem): Tes
 
 function updateTestItem(testItem: TestItem, metaInfo: IJavaTestItem): void {
     testItem.range = asRange(metaInfo.range);
+    testItem.label = metaInfo.label;
```
fladdimir, line 120: > When I change the parameter value and rerun the invocation, the display name won't be updated.

This is should be fixed now by also updating the item.label

src/controller/testController.ts
```diff
@@ -265,23 +265,90 @@ function enqueueTestMethods(testItems: TestItem[], run: TestRun): void {
...
+        if (isInvocation(item)) {
+            dataCache.get(item.parent!)!.uniqueId = dataCache.get(item)!.uniqueId;
+            return item.parent;
+        }
+        return item;
+    })
+
+    removeNonRerunTestInvocations(testItems);
+    return testItems;
+}
```
fladdimir, line 332: > Another issue I found is after running a parameterized tests, return it from the method level will throw an error.
This is because VS Code will give us the method item and all its invocation nodes. Seems that we need to try to merge the test items before the statement const numInvocations: number = testItems.filter((item: TestItem) => dataCache.get(item)?.testLevel === TestLevel.Invocation).length;.

@jdneo The selected invocations are now removed normally in case they are included in another selected higher-level item (e.g. when run from method code lens):
run_all_from_method_code_lens

Also it is now possible to re-run invocations when they have different parents:
rerun_with_different_parents
 
Mixing invocation re-run and method run is also handled:
rerun_with_and_without_parent

Re-running two invocations from the same method will cause the error:
rerun_multiple_error

jdneo (approved):
Change looks great. Thank you @fladdimir for empowering this extension. :)

jdneo:
@fladdimir Thank you for the contribution. The PR is merge. And should be available in the next pre-release (tomorrow).

Issue #1411, opened 2022-04-21
Bug: Re-running multiple invocations of parameterized tests from different methods may run the complete classes instead


fladdimir:
When re-running multiple invocations of different JUnit5 parameterized test methods, not only the selected but all invocations are re-run, in case invocations from all methods of a test class are selected.

Example:
![rerun_bug_class_instead_of_invocations](https://user-images.githubusercontent.com/53265832/164525811-f56a730a-47d0-4672-acb7-f4b00ee9fb05.gif)
-> tests are executed with all parameters (2 stops per method breakpoint); expected behavior: only the parameter-sets of the selected invocations are used (1 stop per test method)

Version: v0.34.2022042102

Related to: PR https://github.com/microsoft/vscode-java-test/pull/1407, Issue https://github.com/microsoft/vscode-java-test/issues/1408

fladdimir:
The problem seems to occur because the current logic always prefers to run a complete test class instead of the single methods, in case all methods of a class are selected:
https://github.com/microsoft/vscode-java-test/blob/2da9eda57016a623acb3ea0782c1a003ee55f98a/src/controller/testController.ts#L449-L456

However, now that a parameterized test method may be restricted to re-run a single parameter-set, there might be the need for another check, since the restriction to single parameter-sets may not be applied when running the complete class?

PR #1412, opened 2022-04-21
fix: re-run invocations from all methods in class

fladdimir:
Only run class instead of methods if no method is restricted to a single invocation.

The example from the issue may now be debugged as expected (no more execution of all parameter-sets):
rerun_bug_fixed

resolve https://github.com/microsoft/vscode-java-test/issues/1411


jdneo (approved):
Thank you, @fladdimirf, then change looks good to me. Maybe we can also try to merge the test invocations when all invocations are selected under a parameterized test method? (Nice to have, not high priority).

Issue #1416, opened 2022-04-26
Stop raising an error when re-running all invocations of a JUnit5 test method


fladdimir:
https://github.com/microsoft/vscode-java-test/pull/1412#pullrequestreview-949352726
Re-running invocations is currently limited to selecting a single invocation, since it works via restricting the parent-method to a single parameter-set.
So when selecting multiple invocations of the same method for a re-run, an error is thrown:

![rerun_all_invocations_before_not_supported](https://user-images.githubusercontent.com/53265832/165363310-5ea46079-1bed-401e-b2c5-4e52d21ecc71.gif)

However, when selecting all invocations, the complete parent method could be re-run without any problem.

jdneo:
To verifier:

1. Prepare a JUnit 5 parameterized test method: [sample](https://github.com/junit-team/junit5-samples/blob/c474e4143775e2f31e23c308638638e24f6b760d/junit5-jupiter-starter-maven/src/test/java/com/example/project/CalculatorTests.java#L29-L40). and run it.
2. In Test explorer, select all the children of `add(int, int, int)`
3. Click Run/Debug Test, 
4. Verify that the tests can be run

PR #1417, opened 2022-04-26
no more error on all-invocations re-run

fladdimir:
As suggested in https://github.com/microsoft/vscode-java-test/pull/1412#pullrequestreview-949352726 , selecting all invocations of a method for re-run does not lead to a "not supported error" anymore. Instead, the method is run normally:

rerun_all_invocations_of_test_method

resolve https://github.com/microsoft/vscode-java-test/issues/1416

### DIFF
--- a/src/controller/testController.ts
+++ b/src/controller/testController.ts
@@ -300,9 +300,6 @@ function handleInvocations(testItems: TestItem[]): TestItem[] {
         return [];
     }
 
-    // remove invocations if they are already included in selected higher-level tests
-    testItems = testItems.filter((item: TestItem) => !(isInvocation(item) && isAncestorIncluded(item, testItems)));
-
     testItems = mergeInvocations(testItems);
 
     const invocations: TestItem[] = filterInvocations(testItems);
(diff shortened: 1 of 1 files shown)


jdneo:
Thank you! I'll start review the change after 0.35.0 is released.

fladdimir:
> Thank you! I'll start review the change after 0.35.0 is released.

Sounds good, thanks!

src/controller/testController.ts
```diff
@@ -311,12 +303,22 @@ function handleInvocations(testItems: TestItem[]): TestItem[] {
     // remove invocations if they are already included in selected higher-level tests
     testItems = testItems.filter((item: TestItem) => !(isInvocation(item) && isAncestorIncluded(item, testItems)));
```
jdneo, line 304: Maybe put this line into the `mergeInvocations()` as well? Since it's dosing a kind of merging operations

jdneo (commented):
Sorry for being late this time. I got a fever and was keeping rest days before. :)

Overall, the change looks good to me. Just a small comment, after that it should be good to be merged.

One more idea is that if we can add some tests for those utils. Do you have interest for doing that one? If not it's totally fine, I can do that later.

fladdimir:
> Sorry for being late this time. I got a fever and was keeping rest days before. :)
> 
> Overall, the change looks good to me. Just a small comment, after that it should be good to be merged.
> 
> One more idea is that if we can add some tests for those utils. Do you have interest for doing that one? If not it's totally fine, I can do that later.

No worries at all, hope you are getting better!

As for the comment and adding some tests, I'd like to do it and I will find some time during the next days.

jdneo:
Thank you @fladdimir! And you can address the comment and adding test in separate PRs. Since we have the pre-release channel. So, it's fine to roll out this fix first.

jdneo (approved):
LGTM. Thank you @fladdimir!

PR #1418, opened 2022-05-04
add some unit-tests for invocations re-run

fladdimir:
The part of the `testController` which handles the preparation for re-running single invocations of parameterized methods was not tested, as discussed in https://github.com/microsoft/vscode-java-test/pull/1417#pullrequestreview-958553125

### DIFF
--- a/src/controller/testController.ts
+++ b/src/controller/testController.ts
@@ -290,7 +290,7 @@ async function getIncludedItems(request: TestRunRequest, token?: CancellationTok
  * @param testItems
  * @returns prepared testItems
  */
-function handleInvocations(testItems: TestItem[]): TestItem[] {
+export function handleInvocations(testItems: TestItem[]): TestItem[] { // export for unit test
 
     if (filterInvocations(testItems)
         .some((invocation: TestItem) => !invocation.parent || !dataCache.get(invocation.parent))) { // sanity-checks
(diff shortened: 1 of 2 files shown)


fladdimir:
@jdneo an attempt for adding some unit-tests of the added preprocessing logic.  
Unfortunately, there does not seem to be a possibility to run unit-tests (for modules that require `'vscode'`) without starting the extension development host?

Does the level make sense for you? Or would it be more useful to go for some higher-level tests against actually exported functions (`runTests`) instead?

src/controller/testController.ts
```diff
@@ -290,7 +290,7 @@ async function getIncludedItems(request: TestRunRequest, token?: CancellationTok
  * @param testItems
  * @returns prepared testItems
  */
-function handleInvocations(testItems: TestItem[]): TestItem[] {
+export function handleInvocations(testItems: TestItem[]): TestItem[] { // export for unit test
```
fladdimir, line 293: may be replaced by using [rewire](https://github.com/jhnns/rewire) ?

jdneo: Using `export` looks good to me now. We can consider using `rewire` if more cases are found.

BTW, we can add the comment `only export for unit test` to the TSDoc, so anyone tries to call it will see that.

jdneo (approved):
Looks great! Thank you @fladdimir for the good job!

PR #1449, opened 2022-07-12
feat: Add codicon in the labels of test items

jdneo:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -323,9 +323,9 @@
             "dev": true
         },
         "@types/vscode": {
-            "version": "1.65.0",
-            "resolved": "https://registry.npmjs.org/@types/vscode/-/vscode-1.65.0.tgz",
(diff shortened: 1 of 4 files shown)


jdneo:
Render result:

#### When no results available
##### New
new_no_result

##### Old
old_no_result

#### Results available
##### New
new

##### Old
old

Eskibear (approved):
looks great

jdneo:
Not sure if users would feel that the UI looks a little bit complex after adding the codicons.

Anyway, let's merge it first and see if we can get some feedbacks from pre-release channel. :)

Eskibear:
I myself don't feel crowded at all. Because the codicon icons can provide further information of the node type, i.e. package/type/method, that I can easily tell the levels...

matter-it-does:
#1453

PR #1459, opened 2022-07-28
add proper error reporting for old testng

gayanper:
TestNG version prior to like 6.7.x didn't failed the test method when
related configuration method failed. Instead they skipped all the
test methods which results in no error message in the Test Explorer.

This change add those missing information if the test method is skipped
followed by a configuration method failure.


jdneo (approved):
The fix works properly when exception happens at methods with TestNG configuration annotations.

Thank you for your contribution @gayanper!

PR #1476, opened 2022-08-25
fix: Update the reference view extension id

### LINKED ISSUE #1475: [Adoption Required] Use `vscode.extensions.getExtension('vscode.references-view')`
VS Code Team here 👋 

In June we have renamed the references viewlet extension. Its NEW name identifier is `vscode.references-view` (old was `ms-vscode.references-view`, different publisher). This extension still uses the old form when calling the `getExtension` API. With the September milestone and release this will NOT work anymore and the new identifier must be used, until then both variants will work. See https://github.com/microsoft/vscode/issues/153320 and https://github.com/microsoft/vscode/issues/152834 for context and details

Let us know if you have follow-up question. 

Happy Coding

jdneo:
fix #1475

### DIFF
--- a/src/commands/navigation/navigationCommands.ts
+++ b/src/commands/navigation/navigationCommands.ts
@@ -11,7 +11,7 @@ import { TestNavigationInput } from './testNavigationInput';
 
 const GENERATE_TESTS: string = 'Generate tests...';
 const SEARCH_FILES: string = 'Search files...';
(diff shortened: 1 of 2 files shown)


Eskibear:
Follow https://github.com/redhat-developer/vscode-java/pull/2517 for backward compatibility, in case it's running on old version of vscode.

Eskibear (approved):

PR #1512, opened 2022-12-13
chore: Update dependencies

jdneo:


java-extension/com.microsoft.java.test.runner/src/main/java/com/microsoft/java/test/runner/testng/TestNGListener.java
```diff
@@ -81,16 +79,6 @@ public void onStart(ISuite suite) {
 
     @Override
     public void onFinish(ISuite suite) {
-        final ITestContext context = getFirst(suite.getResults().values()).getTestContext(); // Can only be one
```
Eskibear, line 84: sure about the changes that you don't need to handle it any more?

jdneo: No, it's redundant now. The VS Code new testing API will handle the test summary. I should delete it long ago when first onboard the new API.

Eskibear (approved):

Issue #1465, opened 2022-08-03
parameterized tests: sorting should be natural instead of alphabetical


matter-it-does:
Make 15  parameterized test cases

https://en.wikipedia.org/wiki/Natural_sort_order

<img width="33" alt="image" src="https://user-images.githubusercontent.com/76812421/182598337-c11a6858-696a-43c5-bc76-24c62bd609cf.png">

jdneo:
Related code is:

https://github.com/microsoft/vscode-java-test/blob/794227efdfe574e4d290db636df7cbd1c57b0786/src/controller/utils.ts#L162-L164

PR #1515, opened 2022-12-21
Fixed how parameterized test methods are sorted

Kropie:
Fixed how parameterized test methods are sorted to ensure that they use a natural sort.

### DIFF
--- a/src/controller/utils.ts
+++ b/src/controller/utils.ts
@@ -160,7 +160,7 @@ export function createTestItem(metaInfo: IJavaTestItem, parent?: TestItem): Test
     }
 
     if (metaInfo.testLevel === TestLevel.Invocation) {
-        item.sortText = metaInfo.sortText;
+        item.sortText = metaInfo.sortText ?? metaInfo.id;
     }
 
     if (parent) {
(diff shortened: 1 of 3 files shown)


Kropie:
This PR will address issue #1465. See the attached screenshot to see the fix in action:
image

jdneo:
Thank you @Kropie!  I'll take a look recently.

Meanwhile, could you fix the linting errors first? You can run `npm run lint` to find out where are the errors.

src/controller/utils.ts
```diff
@@ -160,7 +160,7 @@ export function createTestItem(metaInfo: IJavaTestItem, parent?: TestItem): Test
     }
 
     if (metaInfo.testLevel === TestLevel.Invocation) {
-        item.sortText = metaInfo.id;
+        item.sortText = metaInfo.sortText;
```
jdneo, line 163: Let's fallback to `metaInfo.id` when `metaInfo.sortText` is undefined

jdneo (changes requested):
Looks good overall, just one small change request.

src/controller/utils.ts
Kropie: Sounds good! And I'll look into the linting errors asap

jdneo (approved):
Thank you @Kropie for your contribution!

PR #1532, opened 2023-02-17
fix: Handle methods which have more than 2 arguments

### LINKED ISSUE #1531: Regression: test detection broken, test status broken
Maybe there's a simple repro but this will do.

test is duplicated, one of them doesn't have action buttons. After run, status is not updated.

Pre release version: v0.37.2023021404

suspect https://github.com/microsoft/vscode-java-test/issues/1517 

image

```
package com.blah;

import java.util.stream.Stream;

import org.junit.jupiter.api.Nested;
import org.junit.jupiter.api.Test;
import org.junit.jupiter.api.TestInstance;
import org.junit.jupiter.api.extension.ExtendWith;
import org.junit.jupiter.params.ParameterizedTest;
import org.junit.jupiter.params.provider.Arguments;
import org.junit.jupiter.params.provider.MethodSource;

import io.vertx.junit5.VertxExtension;
import io.vertx.junit5.VertxTestContext;

@ExtendWith(VertxExtension.class)
public class foo {
    @Nested
    @TestInstance(TestInstance.Lifecycle.PER_CLASS)
    class Bar{ Stream<Arguments> data() {
        return Stream.of(
                Arguments.of("foo", "bar"),
                Arguments.of("baz", "qux")
        );
    }
    
    @ParameterizedTest
    @MethodSource("data")
    public void test(String a, String b, VertxTestContext testContext) {
        System.out.println("Hello World");
        testContext.completeNow();
    }
    }
}
```

jdneo:
- Use global match when parsing the method parameters list.
- Eliminate all spaces in the parameters list of the test IDs.

fix #1531


java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestItemUtils.java
```diff
@@ -52,7 +52,8 @@ public static String parseFullName(IJavaElement element, TestLevel level, TestKi
                     // Generics don't come through in the test results, so we need to strip
                     // them out now.
                     final String methodName = JavaElementLabels.getElementLabel(element, JavaElementLabels.ALL_DEFAULT)
-                            .replaceAll("<.*?>", "");
+                            .replaceAll("<.*?>", "")
+                            .replaceAll(" ", "");
```
Eskibear, line 56: is tab legal? how about using \\s ?

jdneo: Take a rough look into the upstream implementation, only ` ` is used. Should be safe here only matching spaces

Eskibear (approved):

matter-it-does:
there's some improvement. but a test like
```
@Test
  void testfoo(Vertx vertx, VertxTestContext vertxTestContext) {
```
still shows up as double.

jdneo:
@matter-it-does

Please file a new issue with some detailed reproduce steps.

matter-it-does:
My bad commented in PR accidentally.

Issue #1534, opened 2023-02-28
Cannot run normal JUnit 5 test methods with multiple parameters


jdneo:
<img width="1210" alt="image" src="https://user-images.githubusercontent.com/76812421/221555543-d2cf135d-4aaf-45b2-bd9d-de0334b85280.png">

```
package com.example.starter;


import io.vertx.core.Vertx;
import io.vertx.junit5.VertxExtension;
import io.vertx.junit5.VertxTestContext;
import org.junit.jupiter.api.Test;
import org.junit.jupiter.api.extension.ExtendWith;


@ExtendWith(VertxExtension.class)
public class TestMainVerticle {

  
  @Test
  void tesfoo(Vertx vertx, VertxTestContext testContext) throws InterruptedException {
    Thread.sleep(1000);
    testContext.completeNow();
  }

}

```

_Originally posted by @matter-it-does in https://github.com/microsoft/vscode-java-test/issues/1531#issuecomment-1446185834_

PR #1535, opened 2023-02-28
fix: Cannot run normal JUnit 5 test methods with multiple parameters

jdneo:
fix https://github.com/microsoft/vscode-java-test/issues/1534


Eskibear (approved):

hotnani (commented):
고

PR #1541, opened 2023-03-01
Adding support for older testng versions

### LINKED ISSUE #1540: New version of test runner doesn't work with older versions of testng
The `getParameterTypes` method is not available in testng before version 7.5 and as such the extension cannot run testng tests older than version 7.5 until the usage of this method is replaced with a suitable alternative.

Kropie:
This will fix #1540

### DIFF
--- a/test/test-projects/old-testng/pom.xml
+++ b/test/test-projects/old-testng/pom.xml
@@ -1,0 +1,18 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>test.project</groupId>
+  <artifactId>test</artifactId>
+  <packaging>jar</packaging>
+  <version>1.0-SNAPSHOT</version>
+  <name>test</name>
+  <url>http://maven.apache.org</url>
+  <dependencies>
+    <dependency>
+      <groupId>org.testng</groupId>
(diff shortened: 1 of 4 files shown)


test/test-projects/old-testng/pom.xml
```diff
@@ -0,0 +1,18 @@
...
+  <artifactId>test</artifactId>
+  <packaging>jar</packaging>
+  <version>1.0-SNAPSHOT</version>
+  <name>test</name>
+  <url>http://maven.apache.org</url>
+  <dependencies>
+    <dependency>
+      <groupId>org.testng</groupId>
+      <artifactId>testng</artifactId>
+      <version>6.10</version>
```
jdneo, line 14: Let's use `6.9.13.3`, since that's the lowest version we declare to support.

jdneo:
I've update the testng version in the test project to `6.9.13.3`.

The PR looks good to me. I'll merge it and arrange a hotfix release.

I'll check if it is possible to do some automation tests towards different versions of the test framework. Let me know if you have any good idea about that :)

test/test-projects/old-testng/pom.xml
jdneo: Done

jdneo (approved):

Kropie:
> I've update the testng version in the test project to `6.9.13.3`.
> 
> 
> 
> The PR looks good to me. I'll merge it and arrange a hotfix release.
> 
> 
> 
> I'll check if it is possible to do some automation tests towards different versions of the test framework. Let me know if you have any good idea about that :)

You could do some simple e2e tests to check that our tests in the various test projects work as expected.

jdneo:
> I am busy for the next few days but I could try to come up with something next week.

It's totally fine. Thank you for all your contributions :)

PR #1549, opened 2023-03-20
chore: Update dependencies

Frederick888:
Failed to build due to 404s. Bumping the dependencies to resolve it.


jdneo (approved):
Thank you @Frederick888. Yes we should keep the file updated as the one declared in JDT.LS: https://github.com/eclipse/eclipse.jdt.ls/blob/master/org.eclipse.jdt.ls.target/org.eclipse.jdt.ls.tp.target.

Thank you for your contribution!

PR #1602, opened 2023-09-02
Add support to specify java executable to use for test

gayanper:
This PR adds support for specifying a different JRE (by providing the java executable) to be used for running the test. This helps when some tests needs to be executed against a different java version to

- Make sure certain application components are future proof.
- Debug issues which has raised when application was running with a newer java version, Which a test is used to reproduce that error.
- etc.

### DIFF
--- a/src/controller/testController.ts
+++ b/src/controller/testController.ts
@@ -225,7 +225,7 @@ export const runTests: (request: TestRunRequest, option: IRunOption) => any = in
                         }
                         try {
                             await runner.setup();
-                            const resolvedConfiguration: DebugConfiguration = option.launchConfiguration ?? await resolveLaunchConfigurationForRunner(runner, testContext, config);
+                            const resolvedConfiguration: DebugConfiguration = mergeConfigurations(option.launchConfiguration, config) ?? await resolveLaunchConfigurationForRunner(runner, testContext, config);
                             resolvedConfiguration.__progressId = option.progressReporter?.getId();
                             delegatedToDebugger = true;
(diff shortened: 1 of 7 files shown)


src/controller/testController.ts
```diff
@@ -225,7 +225,7 @@ export const runTests: (request: TestRunRequest, option: IRunOption) => any = in
                         }
                         try {
                             await runner.setup();
-                            const resolvedConfiguration: DebugConfiguration = option.launchConfiguration ?? await resolveLaunchConfigurationForRunner(runner, testContext, config);
+                            const resolvedConfiguration: DebugConfiguration = mergeConfigurations(option.launchConfiguration, config) ?? await resolveLaunchConfigurationForRunner(runner, testContext, config);
```
jdneo, line 228: Overall looks good to me. Just here:

Previously, if the launch config is passed from other extension, like PDE, we choice not to touch it and directly pass it to debugger.

Here we add a 'merge' logic. Did you meet any issue with the previous implementation?

gayanper:
Yes, this change was initiated for a scenario where I wanted to run some JDT debug tests with java21. This merge logic helps to resolve that. To be safe, i only merge keys which are not present in the config that passed in from PDE, but if we see a need we can extend to concat and merge existing attributes such as strings and arrays.

jdneo:
Ok, makes sense.

jdneo (approved):
LGTM

jdneo:
Thank you @gayanper for the contribution.

PR #1638, opened 2023-12-11
feature: Filter configurations available for tests on launch by pattern

### LINKED ISSUE #1627: Filter by test pattern in config?
Is there a way to control what configurations are available to select when running tests in a specific package, class or method? I am aware there is a `filters` property, but as far as I can see this only supports JUnit tags (`filters.tags`) at the moment. I want to be able to filter by patterns (inspired by the [`Maven Surefire Plugin`](https://maven.apache.org/surefire/maven-surefire-plugin/examples/inclusion-exclusion.html)) - something like this:

`settings.json`

```json
{
    "java.test.config": [
        {
            "name": "tests in specific package"
            "filters": {
                "pattern": "com.example.test"
            }
        },
        {
            "name": "tests not in specific package"
            "filters": {
                "pattern": "!com.example.test"
            }
        },
        {
            "name": "tests in specific class"
            "filters": {
                "pattern": "com.example.test.ExampleTests"
            }
        },
        {
            "name": "tests not in specific class"
            "filters": {
                "pattern": "!com.example.test.ExampleTests"
            }
        },
        {
            "name": "tests in a class of this name"
            "filters": {
                "pattern": "testExample"
            }
        },
        {
            "name": "not tests in a class of this name"
            "filters": {
                "pattern": "!ExampleTests"
            }
        },
        {
            "name": "this specific test"
            "filters": {
                "pattern": "com.example.test.ExampleTests#testExample"
            }
        },
        {
            "name": "not this specific test"
            "filters": {
                "pattern": "!com.example.test.ExampleTests#testExample"
            }
        }
    ]
}
```

ReubenFrankel:
- [x] Implement feature
- [x] Add tests

---

Closes #1627


ReubenFrankel:
@microsoft-github-policy-service agree

jdneo:
Hi @ReubenFrankel,

Thank you for your contribution!

I have one concern about comparing with ID. Since that a kind of internal property and might be changed in the future.

It would be better to compare with something that users are aware of and cleared defined in some specs. Maybe like, a fully qualified name?

ReubenFrankel:
> Hi @ReubenFrankel,
> 
> Thank you for your contribution!
> 
> I have one concern about comparing with ID. Since that a kind of internal property and might be changed in the future.
> 
> It would be better to compare with something that users are aware of and cleared defined in some specs. Maybe like, a fully qualified name?

As far as I can see `TestItem.id` does currently contain the fully-qualified name of the test item, but if that's subject to change then I agree that it would need to be resolved elsewhere.

ReubenFrankel:
@jdneo How's this? https://github.com/microsoft/vscode-java-test/pull/1638/commits/409f6d47f676628b57f94b6985ee193cd79bfa2a#diff-f624d2810b538e481c3a1cd552f8808c287f2ca8c481c00b9b2f19219681b125

jdneo:
Thank you @ReubenFrankel for quick response! I will take a look soon.

jdneo:
Some questions:

1. Looks like the current implementation focus on `include`. While in the original issue, there are some use cases like `!xxx`. I guess you want to use `!` to represent exclusion. Is that request still valid?
2. About the setting description. Since fully qualified name is a clearly defined concept for package and /type. then what about other level? Like `project` and `method`?

It will become complex for methods because JUnit 5 and TestNG allows methods to override. Below is how the extension constructs the `fullName` for methods: https://github.com/microsoft/vscode-java-test/blob/22a11be0c6074c536e40b7b2e4c8fdb05c8803dd/java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestItemUtils.java#L43

I checked the document of [Surefire Plugin](https://maven.apache.org/surefire/maven-surefire-plugin/examples/inclusion-exclusion.html), looks like it does not support methods filtering?

ReubenFrankel:
> 1. Looks like the current implementation focus on `include`. While in the original issue, there are some use cases like `!xxx`. I guess you want to use `!` to represent exclusion. Is that request still valid?

My personal use-case for this feature only really requires inclusion (i.e. I only want a specific configuration to be available when I'm running tests from a specific class). I'm sure there is a case for exclusion, in which case it might make more sense to implement something like:

```json
{
    "filters": {
        "pattern": {
            "include": ...
            "exclude": ...
         }
    }
}
```

The reason I chose regex in the end here is:

1. I'm not sure of the full Surefire plugin spec for test filtering, beyond what I use currently (hence why I also don't know about it not supporting method filtering)
2. It seemed like a heavy lift to support the full Surefire spec or design a custom filtering expression spec
3. Regex is flexible and more widely known
4. You don't need a deep understanding of regex to use `pattern` and can write an expression that matches a class, method or method of a class without regex syntax (as long as the structure of fully-qualfiied names for test items is known):

```json
{
    "filters": {
        "pattern": "TestClass"
    }
}
```

```json
{
    "filters": {
        "pattern": "testMethod"
    }
}
```

```json
{
    "filters": {
        "pattern": "TestClass#testMethod"
    }
}
```

> 2. About the setting description. Since fully qualified name is a clearly defined concept for package and /type. then what about other level? Like `project` and `method`?

A fully-qualified method name takes the pattern `com.package.test.TestClass#testMethod`, as defined [here](https://github.com/microsoft/vscode-java-test/blob/22a11be0c6074c536e40b7b2e4c8fdb05c8803dd/java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/util/TestItemUtils.java#L58). I'm not sure about project... To me, it feels like an edge case but I don't know how many people have configurations that they want to use across multiple projects in the same workspace.

> It will become complex for methods because JUnit 5 and TestNG allows methods to override.

I don't think this changes anything? Overridden methods will have the same name and signature,so just the class name changes, which is easy enough to match:

```json
{
    "filters": {
        "pattern": "(TestClass|AnotherTestClass)#testMethod"
    }
}
```

> For JUnit5 and TestNG, the name of the test method contains the parameter list, while for JUnit 4, it's only the name of the method.

JUnit4
```json
{
    "filters": {
        "pattern": "TestClass#testMethod"
    }
}
```

JUnit5/TestNG
```json
{
    "filters": {
        "pattern": "TestClass#testMethod\(String testArgument\)"
    }
}
```

jdneo:
Thank you @ReubenFrankel. All of your comments make sense to me.

Then let's discuss some details about the implementation.

1. Should we go with

```
"pattern": {
   "include": [...]
   "exclude": [...]
}
```

or

```
"pattern": ["...", "!..."]
```

Considering the tag filtering already adopts the second style, we can adopt it for the pattern filter to make all filter settings align. WDYT?

2. For project filtering: After thinking it again, I now think filtering on project level may not be a real requirement in the real world. And user and just simply click the run test button at the project level in the testing explorer. So we can just simply declare project filtering is not supported.

ReubenFrankel:
> ```
> "pattern": ["...", "!..."]
> ```

I'm not sure this can work with regex, since it will just treat the `!` as a literal character. To work with that, it would have to be a simple string includes/excludes substring operation, which having just thought about it might be fine (regex could very well be overkill here):

```
pattern.startsWith('!') ? !fullName.includes(pattern.slice(1)) : fullName.includes(pattern)
```

I take it this means you want to support multiple patterns for a single configuration, which I hadn't considered but also makes sense.

jdneo:
> I'm not sure this can work with regex, since it will just treat the ! as a literal character.

So, if the first character of the pattern is `!`, we slice it and treat the left part as the pattern for exclusion.

> pattern doesn't really make sense any more if this becomes a substring match.

Personally, I think using regular expression is a good idea. Because IDEA also support that, and the name of it is exactly `pattrn`: https://www.jetbrains.com/help/idea/run-debug-configuration-junit.html

> How do you imagine a glob pattern working for a fully-qualified name?

It's not for fully qualified name, but for the file names. Because in the Surefire Plugin's document, I see usages like:

```
<configuration>
    <excludes>
        <exclude>**/TestCircle.java</exclude>
        <exclude>**/TestSquare.java</exclude>
    </excludes>
</configuration>
```

ReubenFrankel:
> So, if the first character of the pattern is !, we slice it and treat the left part as the pattern for exclusion.

Ok, that can work. Regardless, it just needs documenting well.

> It's not for fully qualified name, but for the file names.

That to me sounds like a separate feature to filter by file path then, in which case glob patterns could be supported of course. Not sure it's strictly related to this PR, although it does raise the question of when/why would a user need to filter by file path if they can filter by pattern already? Something to think about...

jdneo:
> although it does raise the question of when/why would a user need to filter by file path if they can filter by pattern already?

Good ask. To me it looks like pattern is enough. (Guess that's why IDEA only supports that)

I don't know why the Surefire plugin supports both. But since there is a reference from IDEA, it makes me feel safer to move forward.

To summary:

- let's use `pattern` as the setting name
- It's a regular expression match. If the value starts with `!`, it means exclusion.
- The target to match is the fully qualified name. Project level is not taken into consideration.

Sorry for having such a long discussion back and forth. Thanks for the patience. 👍

ReubenFrankel:
> - It's a regular expression match. If the value starts with !, it means exclusion.

To clarify terminology here, is this more like inversion since we are dealing with a pattern, which can both include and exclude (without a `!` prefix) test items depending on the match result? For example, it would be possible for a user to specify some identifier that doesn't exist in a pattern prefixed with `!` to have it match all test items (just demonstrating behaviour - not sure why you would want to do this):

```json
 {
    "filters": {
        "pattern": "!thisdoesntexist"
    }
}
```

> Sorry for having such a long discussion back and forth. Thanks for the patience. 👍

No worries! :slightly_smiling_face: I find this type of design discussion interesting! :sweat_smile:

ReubenFrankel:
By the way, did you want to support multiple patterns for this feature? If so, I have a couple more implementation questions:
- Rename `pattern` to `patterns`?
- Support both `string` and `string[]` in config?
- Every pattern of a configuration must match every test item to be considered a candidate (current implementation is [pattern must match every test item](https://github.com/microsoft/vscode-java-test/blob/d3d224cc96732396e98d2698cb937c99661f1eea/src/utils/configUtils.ts#L55))?

I don't personally have a use-case for multiple patterns at the moment, especially since a you can achieve simple conditional matching in a single regex pattern already:

```json
{
    "filters": {
        "pattern": "(TestClass|AnotherTestClass)"
    }
}
```

I do see it being useful in a couple of situations - for example, match every test in a class apart from one (although this might be possible still in a single pattern with a negative lookahead, albeit not the easiest expression to write):

```json
{
    "filters": {
        "patterns": [
            "TestClass",
            "!TestClass#testMethod"
        ]
    }
}
```

jdneo:
> Rename pattern to patterns?
> Support both string and string[] in config?

Let use `patterns` and make it a `string[]`. If user only has one pattern, just simply use an array with one element: `[".*FooTest"]`.

> Every pattern of a configuration must match every test item to be considered a candidate (current implementation is [pattern must match every test item](https://github.com/microsoft/vscode-java-test/blob/d3d224cc96732396e98d2698cb937c99661f1eea/src/utils/configUtils.ts#L55))?

Instead of `every`, let use `any`. I checked the IDEA's behavior, for example, if user writes `.*Validator.*||.*PetController.*`, IDEA will run tests from `ValidatorTests` and `PetControllerTests`.

> is this more like inversion since we are dealing with a pattern, which can both include and exclude (without a ! prefix) test items depending on the match result?
> this might be possible still in a single pattern with a negative lookahead

Thank you for mentioning this! Negative lookahead can used for exclusion. And I noticed that this is exactly what IDEA does. So let 's not bother using `!` for exclusion. It's just a regular expression that user can define inclusion or exclusion (Negative lookahead) on their own.

ReubenFrankel:
> Instead of `every`, let use `any`. I checked the IDEA's behavior, for example, if user writes `.*Validator.*||.*PetController.*`, IDEA will run tests from `ValidatorTests` and `PetControllerTests`.

If a user runs all tests for a package from the test explorer view, each test item (in this case, each class), doesn't it make sense to make the configuration available only if the pattern satisfies all test items? That's what I meant by "every" - you could still define a single pattern that matches multiple items by itself (like in your example).

jdneo:
I think things should work in the way that:

User selects the configuration, the configuration filter the tests to execute.

Not the selected tests determine which configurations should appear.

ReubenFrankel:
> I think things should work in the way that:
> 
> User selects the configuration, the configuration filter the tests to execute.
> 
> Not the selected tests determine which configurations should appear.

That's not what I was intending this PR to address. From the [original issue](#1627):

> Is there a way to control what configurations are available to select when running tests in a specific package, class or method?

image

For the class `CustomRunner`, I only want the "Custom runner" configurations to show.

jdneo:
Ah, sorry I totally misunderstood the intension of the feature request.

It's to filter the configurations to be displayed, not the tests to be executed.

Since both IDEA and Surefire plugin has the capability to configure the pattern filter for test execution, how can we make sure users are aware the purpose of this setting? (At least, it confused me 😢)

What I'm thinking right now is to put it into a new key, not the key `filter`. For example:

```
java.test.config: {
    ...
    when: {
        matchPatterns: []
    }
```

ReubenFrankel:
> Ah, sorry I totally misunderstood the intension of the feature request.
> 
> It's to filter the configurations to be displayed, not the tests to be executed.

Yeah, I probably could have explained better, sorry - I thought we were on the same page! :sweat_smile:

> Or just simply `when: "..."` without children.

I like that this could follow the VS Code when clause convention. We would have to parse the expression ourselves though, unless there's an API for that already...

> java.test.config: {
>     ...
>     when: {
>         matchPatterns: [...]
>     }
> }

I do like how explicit this is vs the when clause, but agreed that it could get complicated as soon as another condition is added.

jdneo:
> I do like how explicit this is vs the when clause, but agreed that it could get complicated as soon as another condition is added.

Let's go for the vscode when clause approach. WDYT?

ReubenFrankel:
Yep, let's do it. :smile:

src/utils/configUtils.ts
```diff
@@ -40,7 +41,25 @@ export async function loadRunConfig(workspaceFolder: WorkspaceFolder): Promise<I
             return defaultConfigs[0];
         }
     }
-    return await selectQuickPick(configItems, workspaceFolder);
+
+    const testItemMatch: RegExp = /testItem ~= \/(?<pattern>.*)\//;
```
jdneo, line 45: How about:

```typescript
    ...
    const candidateConfigItems: IExecutionConfig[] = filterCandidateConfigItems(configItems);
    return await selectQuickPick(candidateConfigItems, workspaceFolder);
}

function filterCandidateConfigItems(configItems: IExecutionConfig[]): IExecutionConfig[] {
    const matchPattern = /(?<contextKey>.+)\s*~=\s*\/(?<pattern>.+)\//;
    return configItems.filter((config: IExecutionConfig) => {
        const whenClause = config.when?.trim();
        if (whenClause) {
            const groups = matchPattern.exec(whenClause)?.groups;
            if (!groups) {
                return true;
            }
            switch (groups.contextKey) {
                case "testItem":
                    return testItemFiltering(testItems, groups.pattern);
                    break;
                default:
                    return true;
            }
        }
        return true;
    });
}

function testItemFiltering(testItems: TestItem[], pattern: string): boolean {
    ...
}

```

If we want to add some unit test cases, we can consider export checkTestItems() and filterCandidateConfigItems()

src/runConfigs.ts
```diff
@@ -102,6 +102,11 @@ export interface IExecutionConfig {
          */
         tags?: string[]
     }
+
+    /**
+     * The when clause for matching tests by to determine if the configuration should be run with.
```
jdneo, line 107: This can be a new feature in next version, let's add `@since 0.41.0`

package.json
```diff
@@ -320,6 +320,11 @@
                                     "default": []
                                 }
                             }
+                        },
+                        "when": {
+                            "type": "string",
+                            "markdownDescription": "%configuration.java.test.config.when.description%",
+                            "default": "testItem ~= //"
```
jdneo, line 327: How about just leave the default value empty `""`

package.nls.json
```diff
@@ -31,6 +31,7 @@
     "configuration.java.test.config.testKind.description": "Specify the targeting test framework for this test configuration. Supported values are `junit`, `testng`.",
     "configuration.java.test.config.filters.description": "Specify the test filters.",
     "configuration.java.test.config.filters.tags.description": "Specify the tags to be included or excluded. \n\nTags having `!` as the prefix will be **excluded**. \n\nNote: This setting **only** takes effect when `testKind` is set to `junit`.",
+    "configuration.java.test.config.when.description": "Specify the [when clause](https://code.visualstudio.com/api/references/when-clause-contexts) for matching tests by to determine if the configuration should be run with.\n\nWhen launching a test that satisfies a single configuration's when clause, it will be run with that configuration. If multiple configurations are satisfied, the user will be prompted to pick which configuration to run with.\n\nWhen launching multiple tests (e.g. for a class or package), a configuration's when clause must be satisfied for **all** tests to be considered.\n\nConfigurations that do not define a when clause will match all tests.",
```
jdneo, line 34: I suggest avoiding mentioning the when-clause from VS Code document. Because the context key listed there are not supported here, and we do not support `and`, `or` now. It may confuse users.

Just simply tells user that `testItem ~= /<regular-expression>/` is the only clause we support right now.

If possible, give some examples will be even better, like this setting:
Screenshot 2023-12-15 at 10 49 35

jdneo (commented):
Thank you @ReubenFrankel, some feedbacks added.

package.nls.json
ReubenFrankel: Do you want examples in the `package.nls.json`, wiki, or both?

src/runConfigs.ts
ReubenFrankel: https://github.com/microsoft/vscode-java-test/pull/1638/commits/338c10dba870cf61cc170143f7b6008d8a0c41b9

package.json
ReubenFrankel: https://github.com/microsoft/vscode-java-test/pull/1638/commits/338c10dba870cf61cc170143f7b6008d8a0c41b9

src/utils/configUtils.ts
ReubenFrankel: https://github.com/microsoft/vscode-java-test/pull/1638/commits/469518512bb286d39bffa91279fab306fcca6c58

src/runConfigs.ts
ReubenFrankel: https://github.com/microsoft/vscode-java-test/pull/1638/commits/751f3d5ab45ebdf17a88712ab3277175b8ee100c

ReubenFrankel: https://github.com/microsoft/vscode-java-test/pull/1638/commits/751f3d5ab45ebdf17a88712ab3277175b8ee100c

ReubenFrankel:
@jdneo I have been working on [a utility class to evaluate a when-clause](https://github.com/ReubenFrankel/vscode-java-test/commit/0e4a48d5cc4c509125b324414a41d852a9d2120b), since there is no equivalent public VS Code API (that I could find). Are you interested in that as part of this feature? Obviously, it will need some unit tests (which I did start writing) and a little more manual testing.

package.nls.json
jdneo: Yes, just give one or two examples will be good enough.

I'll take care of the package.nls.zh.json & wiki.

jdneo:
> @jdneo I have been working on [a utility class to evaluate a when-clause](https://github.com/ReubenFrankel/vscode-java-test/commit/0e4a48d5cc4c509125b324414a41d852a9d2120b), since there is no equivalent public VS Code API (that I could find). Are you interested in that as part of this feature? Obviously, it will need some unit tests (which I did start writing) and a little more manual testing.

Yeah, I like that! Thank you for the effort!

package.nls.json
ReubenFrankel: https://github.com/microsoft/vscode-java-test/pull/1638/commits/9cbdbbdc51ccc8b85bbd5fba272474d44ee34eaf

src/utils/configUtils.ts
```diff
@@ -40,7 +41,38 @@ export async function loadRunConfig(workspaceFolder: WorkspaceFolder): Promise<I
...
+        return true;
+
+    });
+}
+
+function checkTestItems(testItems: TestItem[], context: WhenClauseEvaluationContext): boolean {
+    return testItems.every((testItem: TestItem) => {
+        const fullName: string | undefined = dataCache.get(testItem)?.fullName;
+
+        context.addContextKey('testItem', fullName);
```
jdneo, line 73: this can be moved out of this `every` loop

src/utils/configUtils.ts
```diff
@@ -103,3 +135,149 @@ async function askPreferenceForConfig(configs: IExecutionConfig[], selectedConfi
...
+        // match
+        '=~': (value: string, pattern: RegExp) => pattern.test(value),
+    }
+
+    private readonly context: Record<string, unknown> = {};
+
+    public constructor(readonly clause: string) {}
+
+    private tokenize(): Token[] {
+        const tokens: string[] = this.clause.split(/\s+/)
```
jdneo, line 179: Do you want to support clause without space, like `testItem=~xxx`?

src/utils/configUtils.ts
```diff
@@ -103,3 +135,149 @@ async function askPreferenceForConfig(configs: IExecutionConfig[], selectedConfi
...
+            .flatMap((token: string) => token.split(/([\(\)]|!(?!=))/))
```
jdneo, line 180: I guess `[()]` can be used instead of `[\(\)]`?

src/utils/configUtils.ts
```diff
@@ -103,3 +135,149 @@ async function askPreferenceForConfig(configs: IExecutionConfig[], selectedConfi
...
+        if (regexMatch?.groups) {
+            const { pattern, flags } = regexMatch.groups;
+            return new RegExp(pattern, flags);
+        }
+
+        const number: number = Number(token);
+        if (!isNaN(number))
+            return number;
+
+        const booleanMatch: RegExpMatchArray | null = token.match(/^true|false$/);
```
jdneo, line 204: `/^true|false$/` this will match `^true` or `false$`. I would suggest changing it to `/^(?:true|false)$/`

src/utils/configUtils.ts
```diff
@@ -40,7 +41,38 @@ export async function loadRunConfig(workspaceFolder: WorkspaceFolder): Promise<I
...
+        const whenClause: string | undefined = config.when?.trim();
+
+        if (whenClause) {
+            const context: WhenClauseEvaluationContext = new WhenClauseEvaluationContext(whenClause);
+
+            try {
+                return checkTestItems(testItems, context);
+            } catch (e) {
+                if (e instanceof Error)
+                    window.showWarningMessage(e.message);
```
jdneo, line 60: Usually we should be careful about using the notifications. Because it will grab the user's attention. See: https://code.visualstudio.com/api/ux-guidelines/notifications

Here, I would recommend failing quietly for those configurations have a wrong when clause. (in other word, ignore the when clause if it's not written correctly).

If we want to let user know the when clause is wrong, we can use things like diagnostics: https://github.com/microsoft/vscode-extension-samples/tree/main/diagnostic-related-information-sample.

But it's out of the scope of this PR, so far just ignore the when clause will be good enough to me :)

jdneo (commented):
Overall looks good, added some new comments.

src/utils/configUtils.ts
ReubenFrankel: From https://code.visualstudio.com/api/references/when-clause-contexts#comparison-operators:

> You can compare a context key's value against a number. Note that left- and right-hand side of the operator must be separated by whitespace - `foo < 1`, but not `foo<1`.

It doesn't specify if this is the case for all operators, but all examples given point to that. If not, we could split on all operator keys (and retain them):

```js
        const operatorKeys: string[] = Object.keys(WhenClauseEvaluationContext.OPERATORS);
        const operatorPattern: RegExp = new RegExp(`(${operators.map(_.escapeRegExp).join('|')}`);

        const tokens: string[] = this.clause.split(operatorPattern)
            .flatMap((token: string) => token.split(/([\(\)]|!(?!=))/))
            .filter(Boolean);
```

jdneo: I see. Let's use whitespace as the separator then.

src/utils/configUtils.ts
ReubenFrankel: https://github.com/microsoft/vscode-java-test/pull/1638/commits/b231049b1bd33950a32c3d55c7d89c365f729574

src/utils/configUtils.ts
ReubenFrankel: https://github.com/microsoft/vscode-java-test/pull/1638/commits/d73612c601c048d3398cb98f0b4127473dd14c76

src/utils/configUtils.ts
ReubenFrankel: Sorry, edited my comment above after you had replied. Any other thoughts?

src/utils/configUtils.ts
ReubenFrankel: Fair enough - assume this applies to https://github.com/microsoft/vscode-java-test/blob/b231049b1bd33950a32c3d55c7d89c365f729574/src/utils/configUtils.ts#L208 too? I'll leave the try/catch block there so we can still throw errors from `evaluate` and return the configuration as a candidate.

ReubenFrankel: https://github.com/microsoft/vscode-java-test/pull/1638/commits/df5da6b4d71d294d8496608ab39d432a3ab06be4

src/utils/configUtils.ts
ReubenFrankel: FWIW, I tested the above an it works. Can revert https://github.com/microsoft/vscode-java-test/pull/1638/commits/5c6135e89fa896f168cc58d08a22b71fb5fd669b if not required.

ReubenFrankel:
I guess we can link off to the VS Code reference for the when-clause syntax now? The only features not supported now are the `in` and `not in` operators. These would be easy enough to add but probably not necessary at the moment, since the only context key available for configurations is `testItem`.

src/utils/configUtils.ts
jdneo: It looks good! No need to revert 👍

jdneo:
Appended a commit to add chinese description and did some simplification.

I prefer to progressively adding more description on demand (if users are confused.)

So Let wait and see if we need to link to the VS Code reference for the when-clause syntax.

jdneo (approved):
LGTM!

I'll wait if you have more comments on the appended commit. If not, I'll merge it.

Merry Christmas! 🎉

ReubenFrankel:
Last thing would maybe be to remove the bit of the description about the only supported clause (`=~`) and just mention the only available context key (`testItem`)? Other than that, I'm happy if you're happy!

Merry Christmas! 😁🎄

jdneo:
> Last thing would maybe be to remove the bit of the description about the only supported clause (=~) and just mention the only available context key (testItem)?

I tend to keep as what it is since `testItem =~ xxx` is the only pattern we support to users (though the implementation can do further)

jdneo:
Merged. Thank you @ReubenFrankel for your contribution.

PR #1669, opened 2024-03-01
Update vscode-java-test-nightly.yml for api scan

testforstephen:


jdneo (commented):
The line 106 (replace AI key task) may fail due to the changing of the OS.

```
On UNIX systems (macOS and Linux), environment variables have the format $NAME. On Windows, the format is %NAME% for batch and $env:NAME in PowerShell.
```

_https://learn.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#environment-variables_

jdneo (approved):

PR #1735, opened 2024-09-24
Set which encoding your test JVM will start with

### LINKED ISSUE #1641: "vmArgs": ["-Dfile.encoding=iso885915"] seems to be overided by something else in the java.test.config config
Hello

when I put some specific args in the java.test.config, they are passed to code under test, except file.encoding, which is always stuck to UTF-8

```
"java.test.config": {
        "name": "MonConfig",

        "vmArgs": [
            "-Dfile.encoding=iso8859-15",
            "-DTITI=iso8859-15",
        ],
}
```

==>
```
mainid 2023-12-15 16:27:38,655 - INFO  ApplicationProperties - file.encoding = UTF-8
mainid 2023-12-15 16:27:38,657 - INFO  ApplicationProperties - TITI = iso8859-15
```

It seems that something is overiding this setting

The project is build by gradle. My build.gradle contains something like this :

```
   tasks.withType(Test) {
        // Configuration spécifique pour toutes les tâches de test des sous-projets
        systemProperty 'file.encoding', 'iso8859-15'
   }
```
and it work well when i use directly gradle in my vs console
```
    Test workerid 2023-12-15 16:52:15,934 - INFO  ApplicationProperties - file.encoding = iso8859-15
```

awilkins:
This patch adds a setting to the `java.test.config` object.

```jsonc
{
  "java.test.config": {
    "encoding": "ISO-8859-1",
  },
}
```
The aim here is to be able to seamlessly test code which cares about the launch encoding of the JVM. Yes, such code is not good. But sometimes you don't control your dependencies enough to fix this.

Since the debug launcher already cares about the `encoding` member of the config, the only thing required is to add the member, which is missing from the launch configs generated by the test plugin.

```typescript
// In `configurationProvider.ts:` (vscode-java-debug)
// VS Code internal console uses UTF-8 to display output by default.
if (config.console === "internalConsole" && !config.encoding) {
    config.encoding = "UTF-8";
}
```

Fixes #1641

NB : I couldn't work out how to actually launch and gather results from Java tests, in the Typescript test suite, but I did add a Java test case that cares about the launch encoding and tested it manually.


awilkins:
PS : had to change the update site for eclipse to

```xml
            <repository location="https://download.eclipse.org/eclipse/updates/4.33/R-4.33-202409030240/"/>
```

: the older URL doesn't seem available any more. I'm presuming your builds are behind a Maven repo that has this stuff cached, but building the source as-is from publicly available builds isn't currently possible.

jdneo:
> PS : had to change the update site for eclipse to
> 
> ```
>             <repository location="https://download.eclipse.org/eclipse/updates/4.33/R-4.33-202409030240/"/>
> ```
> 
> : the older URL doesn't seem available any more. I'm presuming your builds are behind a Maven repo that has this stuff cached, but building the source as-is from publicly available builds isn't currently possible.
> 
> I've not submitted this as part of the PR (because I figure you have that Maven repo), but I'm mentioning it so you know.

Updated the target platform definition in https://github.com/microsoft/vscode-java-test/pull/1734

jdneo:
Hi @awilkins,

Sorry that I'm just back from vacation and I will review this ASAP.

package.json
```diff
@@ -254,6 +254,11 @@
                             "markdownDescription": "%configuration.java.test.config.javaExec.description%",
                             "default": ""
                         },
+                        "encoding": {
```
jdneo, line 257: Since `java.test.config` can either be an array or an object. So we need to duplicated this under `item` section as well. (start from line 349)

awilkins: Also noted, because they're right next to each other, that it's `vmargs` in the array and `vmArgs` in the singular ; have changed the case of the array to match (this has been deprecated 4 years, I think it might have been long enough :-) )

jdneo: Thank you

jdneo (approved):
LGTM

jdneo:
Thank you @awilkins

Issue #1796, opened 2025-10-09
Bundle startup failed


ooMia:
The latest [`vscode-java`](https://github.com/redhat-developer/vscode-java) build embeds `org.objectweb.asm 9.9.0`, so `vscode-java-test` extension needs to update its dependency in future release.

> - issue from https://github.com/redhat-developer/vscode-java/issues/4205
>
> ### [Language Support for Java.log](https://github.com/user-attachments/files/22792749/Language.Support.for.Java.log)
>
> I could reproduce, logs show:
> 
> ```
> org.osgi.framework.BundleException: Could not resolve module: com.microsoft.java.test.plugin [117]
>   Unresolved requirement: Require-Bundle: org.objectweb.asm; bundle-version="[9.8.0,9.9.0)"
>   Unresolved requirement: Require-Bundle: junit-jupiter-engine; bundle-version="5.4.0"
>     -> Bundle-SymbolicName: junit-jupiter-engine; bundle-version="5.11.0"
>        junit-jupiter-engine [101]
>          No resolution report for the bundle.  Unresolved requirement: Require-Bundle: org.eclipse.jdt.junit5.runtime
>     -> Bundle-SymbolicName: org.eclipse.jdt.junit5.runtime; bundle-version="1.1.300.v20231214-1952"; singleton:="true"
>        org.eclipse.jdt.junit5.runtime [114]
>          Unresolved requirement: Require-Bundle: junit-jupiter-engine; bundle-version="5.4.0"
>            -> Bundle-SymbolicName: junit-jupiter-engine; bundle-version="5.11.0"
>          Unresolved requirement: Require-Bundle: junit-jupiter-params; bundle-version="5.4.0"
>            -> Bundle-SymbolicName: junit-jupiter-params; bundle-version="5.11.0"
>               junit-jupiter-params [103]
>   Unresolved requirement: Require-Bundle: junit-jupiter-params; bundle-version="5.4.0"
>     -> Bundle-SymbolicName: junit-jupiter-params; bundle-version="5.11.0"
> ```
> 
> The latest vscode-java build embeds org.objectweb.asm 9.9.0, so https://github.com/microsoft/vscode-java-test/blob/e3eb3bd13171cd8775d5cd051907f18d6e9c366d/java-extension/com.microsoft.java.test.plugin/META-INF/MANIFEST.MF#L43 needs to be updated to relax the constraint. The previous release included 9.8.0 so it worked.
> 
[...]

PR #1798, opened 2025-10-13
fix - Update to Jacoco 0.8.14

fbricon:
Fixes https://github.com/microsoft/vscode-java-test/issues/1796

Update to Jacoco 0.8.14, which requires org.objectweb.asm:9.9.0.
The caveat is the extension requires vscode-java 1.47.0 to work, i.e. the current vscode-java pre-release, since jacoco no longer works with asm 9.8.0.


wenytang-ms (approved):
This PR also fix the ASM error in current `npm run build-plugin` issue

dbi1463:
Thanks for fixing this!
Any idea when the release containing this change will go out?

PR #1799, opened 2025-10-13
fix: update parent node state when test completes

fbricon:
Test class/package status doesn't update until the end of the full test run: 
https://github.com/user-attachments/assets/2bc8c942-b407-447b-9c10-9d84c5403040

With this PR, completed tests update their parent node status, making updates/progress less frustrating:
https://github.com/user-attachments/assets/527e83da-0837-44a9-afa3-927aaa035741


wenytang-ms (approved):
LGTM

PR #1859, opened 2026-04-23
Fix JUnit 5 launch failure with standalone/fat Multi-Release JARs

wenytang-ms:
## Summary

Fix JUnit 5 test launch failure when using standalone/fat Multi-Release JARs such as `junit-platform-console-standalone-1.13.4.jar`.

## Problem

Since the upstream Eclipse JDT change ([eclipse-jdt/eclipse.jdt.ui@0edde57](https://github.com/eclipse-jdt/eclipse.jdt.ui/commit/0edde57)), `CoreTestSearchEngine.findAnnotations()` uses `NameLookup.findType()` with `ACCEPT_ANNOTATIONS` + `checkRestrictions=true` instead of the old `IJavaProject.findType()`. This causes `@Testable` to not be found in standalone/fat Multi-Release JARs, resulting in:

`
Cannot find 'org.junit.platform.commons.annotation.Testable' on project build path.
JUnit 5 tests can only be run if JUnit 5 is on the build path.
`

## Fix

Override `preLaunchCheck()` in our `JUnitLaunchConfigurationDelegate` to:
1. Call `super.preLaunchCheck()` normally (preserving all existing validation)
2. Catch only the specific `ERR_JUNIT_NOT_ON_PATH` (20006) error
3. Fall back to `IJavaProject.findType()` to verify JUnit is actually on the classpath
4. Only suppress the error if the fallback confirms JUnit is present; otherwise re-throw

This is a **narrow, targeted workaround** — all other validation from the superclass is preserved.

## Related Issues

- Downstream: https://github.com/redhat-developer/vscode-java/issues/4396
- Upstream bug filed: https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2959

### DIFF
--- a/java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/launchers/JUnitLaunchConfigurationDelegate.java
+++ b/java-extension/com.microsoft.java.test.plugin/src/main/java/com/microsoft/java/test/plugin/launchers/JUnitLaunchConfigurationDelegate.java
@@ -31,6 +31,7 @@ import org.eclipse.jdt.core.ICompilationUnit;
 import org.eclipse.jdt.core.IJavaProject;
 import org.eclipse.jdt.core.IMethod;
 import org.eclipse.jdt.core.JavaCore;
+import org.eclipse.jdt.core.JavaModelException;
 import org.eclipse.jdt.core.dom.CompilationUnit;
 import org.eclipse.jdt.core.dom.ITypeBinding;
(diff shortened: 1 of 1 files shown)


CherryDT:
Why was this closed?

wenytang-ms:
@CherryDT This PR was just a workaround; it didn't solve the real problem and could easily introduce versioning issues. So, I submitted a proper fix, which has now been merged and will be available for verification in the next `vscode-java` release.
https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2960

Issue #1836, opened 2026-02-16
Multiple selections in test UI run individually/isolated


Terrance:
_Originally raised at redhat-developer/vscode-java#4338._

Observed behaviour: multiple methods under a Java test class (or multiple test classes under a package) selected in the _Tests_ sidebar tree seem to run significantly slower than when just selecting the single parent class (or package).

This seems to be because each selection does a full setup/teardown (i.e. selecting a class does one setup/teardown cycle, before the first and after the last method; selecting four methods from that class does four cycles, one around each method).  For projects like Spring web applications and their unit tests, the application is recreated unnecessarily for each test, and for large applications this causes a significant slowdown.  Repeating failed unit tests from a previous run doesn't seem to be affected in this way.

My understanding is that JUnit at least doesn't provide much of a test selection interface, so I'm assuming there's a custom runner in play in order to target the selected or previously-failed tests in the first place.

wenytang-ms:
Hi @Terrance — thanks for the detailed report. I dug into the code path and wanted to share the root cause, because the fix is unfortunately less local than it first appears.
What an actual fix would require

To get a single JVM for an arbitrary multi-method selection we'd have to bypass the stock RemoteTestRunner protocol. The realistic option is to extend the in-tree custom launcher (com.microsoft.java.test.runner-jar-with-dependencies.jar, currently used for TestNG only) with a
JUnit Platform mode that:

 - depends on junit-platform-launcher,
 - builds a single LauncherDiscoveryRequest with multiple DiscoverySelectors.selectMethod(class, method) selectors, and
 - streams events back over the existing socket protocol (mirroring the TestNG launcher).

Routing same-class (and ideally cross-class) multi-method batches through that new launcher in mergeTestMethods() + resolveLaunchConfigurationForRunner() would then collapse them into one JVM. JUnit 4 would still be limited unless we also wire up junit-vintage-engine in that
launcher.

wenytang-ms:
The protocol PR is already merged, need jdtls to ship 3.8.100, then vscode-java-test may follow the new bits to ship the pre-release `vscode-java-test`

PR #1862, opened 2026-05-06
Run multiple selected test methods in a single JVM (#1836)

wenytang-ms:
When several test methods of the same class are selected in the Test Explorer, vscode-java-test currently launches one JVM per method, which makes Spring-style tests (where each JVM rebuilds the ApplicationContext in @BeforeAll) extremely slow.

Group methods of the same class into one launch and rely on the new 'Class:method' line format supported by the bundled Eclipse JDT JUnit runtime so that all selected methods are discovered inside a single JVM, sharing per-class @BeforeAll/@AfterAll lifecycle and any cached fixture (e.g. Spring ApplicationContext).

Methods restricted to a single invocation (uniqueId) keep their own launch, since the underlying protocol carries at most one uniqueId per JVM.


wenytang-ms:
depends on https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2975

chagong (approved):

Issue #1866, opened 2026-05-28
Test Runner for Java Extension downloads old JUnit standalone


joaocaram:
When we create a "no build tools" Java project and, after that, try to "Enable Java Tests" using the "test icon", it downloads the "junit-platform-console-standalone-1.13.0-M3.jar" file instead of JUnit 6 versions. This old version is not working with most recent Test Runner and Language Support for Java Extensions.  (i'm using Extension Pack for Java 0.31.0, Language Support 1.54.0 and Test Runner 0.45.0).

<img width="712" height="526" alt="Image" src="https://github.com/user-attachments/assets/9bc4a34a-2b69-4f10-bd96-3421ec192ffa" />

<img width="432" height="257" alt="Image" src="https://github.com/user-attachments/assets/d8bfcfae-d9eb-450c-9d38-70cde0775e89" />

I know we can fix it by manually downloading more recent standalones, or creating Maven/Gradle projects. But when the platform is being used to teach some dozen first-term students, creating a Maven project, keep in mind the directories and packages structure etc are annoying and time consuming distractions. 

Besides, if there is a plugin to make things easier for everyone (such as downloading and including the reference to the Junit jar), it should work properly. So, please, take a few minutes to fix this.

joaocaram:
Hi AI/bot @github-actions

>     * Upvote or comment on the existing feature request to bump the default standalone JAR version in the Test Runner extension so the next release ships a recent JUnit platform.

Sorry, I may be missing something here. Where do I find the feature request so I can upvote? 

Thanks

wenytang-ms:
Looked into this. The root cause is actually two bugs working together:

1. **`Enable Java Tests` downloads a Platform 1.x jar** ([`testDependenciesCommands.ts:81,138-142`](https://github.com/microsoft/vscode-java-test/blob/main/src/commands/testDependenciesCommands.ts#L81)): the `JUnit5` branch asks Maven Central's search API for `latestVersion` of `org.junit.platform:junit-platform-console-standalone`, which currently returns `1.13.x`.
[...]

PR #1868, opened 2026-06-02
fix: use maven-metadata.xml to look up latest test framework version

wenytang-ms:
## Problem

`search.maven.org`'s Solr index was frozen as part of Sonatype's migration
to the new Central Portal — `migration.json` returns `{"successful":[],
"failed":[]}` for every artifact we ship, so the legacy endpoint silently
stopped returning new releases.

For `junit-platform-console-standalone` that left the extension picking
the milestone build `1.13.0-M3` (last entry in the stale index) instead
of a real stable release, breaking "Enable Java Tests" on unmanaged
folders. See #1866 for the full report.

### UI - User select Test Framework
image

### Switch the data source (fix)

Replace the `search.maven.org` Solr lookup with
`repo1.maven.org/.../maven-metadata.xml`, the Maven 2 metadata file that
every artifact on Maven Central is required to publish and that is always
in sync with what Central actually serves. This removes the dependence on
a deprecated, frozen index.

### Split JUnit Jupiter 5 and JUnit Jupiter 6 by version line

`junit-platform-console-standalone` now publishes two coexisting release
lines under the same GAV — `1.x` for Jupiter 5 and `6.x` for Jupiter 6 —
and `<release>` tracks the globally newest stable (currently `6.1.0`).
Without a per-line constraint, picking *JUnit Jupiter 5* would still
resolve to a 6.x jar.

The QuickPick now exposes two entries:

| Menu label        | Platform line | Default fallback |
|-------------------|---------------|------------------|
| JUnit Jupiter 5   | `1.x`         | `1.14.4`         |
| JUnit Jupiter 6   | `6.x`         | `6.1.0`          |

To support this, `IArtifactMetadata` gains an optional `versionLine`
field. When set, `getLatestVersion()` / `parseLatestStableVersion()`
skip the `<release>` shortcut and scan `<versions>` for the newest
stable version whose **leading segment** matches the requested line
(`'1'` matches `1.14.4` but not `10.x` or `11.x`).

### Add regression coverage

| Layer | File | Catches |
|---|---|---|
| Integration | `test/unmanaged-folder-suite/enableTests.test.ts` | Production code path: tightened to require a stable **1.x** `junit-platform-console-standalone-*.jar` after `enableTests(JUnit5)`, so any silent re-introduction of `-M3` / `-RC1` builds or accidental drift back to the 6.x line fails CI. |
| Unit | `test/suite/testDependenciesCommands.test.ts` | XML parser invariants: `<release>` happy path, fallback to `<versions>` when `<release>` is a pre-release (mirrors slf4j-api shape), and the new per-line cases — selecting newest stable in the requested line, ignoring an out-of-line `<release>`, skipping pre-releases inside the line, returning `undefined` when the line has nothing stable, and using leading-segment match (not prefix). |
| Health check | `scripts/checkVersionSources.js` + `.github/workflows/health-check.yml` | Future Sonatype breakage: weekly cron + PR/push runs that HEAD every artifact we ship (now probes both the 1.x and 6.x console-standalone lines independently) and opens / updates a tracking issue on scheduled failure. |

## Known follow-up: JUnit Jupiter 5 + current vscode-java

There is a separate upstream JDT-LS dispatch bug that routes **every**
`junit-platform-console-standalone` jar through the new "JUnit 6" runner,
which reflectively requires APIs only present in Platform 6+. So on a
vscode-java build that still bundles the affected JDT-LS, picking
*JUnit Jupiter 5* (1.x) will fail with `NoClassDefFoundError:
.../CancellationToken`.

The upstream fix is merged (eclipse.jdt.ui#2910 / `17f8fa5d9f`) and is
waiting on the eclipse.jdt.ls bump + the next `redhat.java` release.
Until then, affected users should pick *JUnit Jupiter 6* or roll back
to an older `redhat.java`. #4396 tracks the rollout.

## Closes
- #1866

### DIFF
--- a/.github/workflows/health-check.yml
+++ b/.github/workflows/health-check.yml
@@ -47,6 +47,7 @@ jobs:
           script: |
             const runUrl = `${context.serverUrl}/${context.repo.owner}/${context.repo.repo}/actions/runs/${context.runId}`;
             const title = '[health-check] Maven Central data source check failed';
+            const labelName = 'ci-health';
             const body = [
               'The scheduled health check for the Maven Central data sources used by',
               '`scripts/checkVersionSources.js` failed.',
@@ -64,11 +65,29 @@ jobs:
               'Please investigate before users start hitting the broken download.',
             ].join('\n');
 
+            // Ensure the label exists so issues.create({ labels: [...] }) does not 422.
+            try {
+              await github.rest.issues.createLabel({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                name: labelName,
+                color: 'd93f0b',
+                description: 'Automated CI health-check failure',
+              });
+              core.info(`Created label "${labelName}".`);
+            } catch (err) {
+              if (err.status === 422) {
+                core.info(`Label "${labelName}" already exists.`);
+              } else {
+                throw err;
+              }
+            }
+
             const { data: existing } = await github.rest.issues.listForRepo({
               owner: context.repo.owner,
               repo: context.repo.repo,
               state: 'open',
-              labels: 'ci-health',
+              labels: labelName,
               per_page: 100,
             });
             const alreadyOpen = existing.find((issue) => issue.title === title);
@@ -86,7 +105,7 @@ jobs:
                 repo: context.repo.repo,
                 title,
                 body,
-                labels: ['ci-health'],
+                labels: [labelName],
               });
               core.info(`Opened issue #${created.number}.`);
             }
--- a/.github/workflows/health-check.yml
+++ b/.github/workflows/health-check.yml
@@ -1,0 +1,92 @@
+name: Health Check - Maven Central Data Sources
+
+# Verifies that the upstream Maven Central data sources used by the
+# "Enable Java Tests" download flow (see scripts/checkVersionSources.js)
+# are still healthy. This catches situations like microsoft/vscode-java-test#1866,
+# where the legacy search.maven.org Solr index was silently frozen and
+# kept returning a pre-release as the "latest" stable version.
+#
+# - Runs on every PR and push to main so a code change cannot break the lookup.
+# - Runs weekly so purely upstream changes are caught within days.
+# - Can be triggered manually from the Actions tab.
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+  schedule:
+    # Monday 09:00 UTC.
+    - cron: '0 9 * * 1'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  issues: write
+
+jobs:
+  check-version-sources:
+    name: Check Maven Central version sources
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Setup Node.js environment
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+
+      - name: Run version source health check
+        run: node scripts/checkVersionSources.js
+
+      - name: Open issue on scheduled failure
+        if: failure() && github.event_name == 'schedule'
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const runUrl = `${context.serverUrl}/${context.repo.owner}/${context.repo.repo}/actions/runs/${context.runId}`;
+            const title = '[health-check] Maven Central data source check failed';
+            const body = [
+              'The scheduled health check for the Maven Central data sources used by',
+              '`scripts/checkVersionSources.js` failed.',
+              '',
+              `Failed run: ${runUrl}`,
+              '',
+              'This typically means one of the following:',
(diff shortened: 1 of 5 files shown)


src/commands/testDependenciesCommands.ts
```diff
@@ -138,7 +138,7 @@ function getJarIds(testKind: TestKind): IArtifactMetadata[] {
             return [{
                 groupId: 'org.junit.platform',
                 artifactId: 'junit-platform-console-standalone',
-                defaultVersion: '1.9.3',
+                defaultVersion: '6.1.0',
```
wenytang-ms, line 152: maven-metadata.xml release config is update to `6.1.0`, just align to the latest version.

src/commands/testDependenciesCommands.ts
```diff
@@ -274,9 +304,9 @@ async function updateProjectSettings(projectUri: Uri, libFolder: string): Promis
     window.showInformationMessage(`Test libraries have been downloaded into '${relativePath}/'.`);
 }
 
-async function getHttpsAsJSON(link: string): Promise<any> {
+async function getHttpsAsText(link: string): Promise<string> {
     // eslint-disable-next-line @typescript-eslint/typedef
-    const response: string = await new Promise<string>((resolve, reject) => {
+    return new Promise<string>((resolve, reject) => {
         let result: string = '';
         https.get(link, {
```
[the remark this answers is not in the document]
wenytang-ms: Good catch — fixed in 68b33a6. `getHttpsAsText` now binds `req.on('error', reject)` so DNS / TCP / TLS failures no longer hang the promise, drains the response body via `res.resume()` before rejecting on non-2xx, and accepts any 2xx status (not just 200).

src/commands/testDependenciesCommands.ts
```diff
@@ -172,20 +172,50 @@ function getJarIds(testKind: TestKind): IArtifactMetadata[] {
...
+        const xml: string = await getHttpsAsText(getMetadataLink(groupId, artifactId));
+        const version: string | undefined = parseLatestStableVersion(xml);
+        if (version === undefined) {
+            sendError(new Error(`No stable version found in maven-metadata.xml for ${groupId}:${artifactId}`));
         }
-        return response.response.docs[0].latestVersion;
+        return version;
     } catch (e) {
         sendError(new Error(`Failed to fetch the latest version for ${groupId}:${artifactId}`));
     }
```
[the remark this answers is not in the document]
wenytang-ms: Fixed in 68b33a6 — the `catch` now extracts `e.message` (or `String(e)` for non-Error throwables) and includes it in the reported error so telemetry preserves the root cause.

.github/workflows/health-check.yml
```diff
@@ -0,0 +1,92 @@
...
+              });
+              core.info(`Commented on existing issue #${alreadyOpen.number}.`);
+            } else {
+              const { data: created } = await github.rest.issues.create({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                title,
+                body,
+                labels: ['ci-health'],
+              });
```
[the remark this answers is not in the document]
wenytang-ms: Good catch — fixed in 68b33a6. The workflow now idempotently creates the `ci-health` label via `issues.createLabel` (treating HTTP 422 `already_exists` as success) before listing / creating the tracking issue, so the scheduled-failure branch no longer 422s on a fresh checkout.

src/commands/testDependenciesCommands.ts
chagong, line 152: any risk on such big bump for the version?

wenytang-ms: `defaultVersion` is the fallback when `getLatestVersion()` fails. `getLatestVersion()` uses the `<release>` directive in `maven-metadata.xml`, and since `<release>` is currently set to 6.1.0, the fallback directive should use 6.1.0 to align with it.

wenytang-ms: 1. "Why change the fallback?" — The fallback was changed because without it, the results in the online environment (6.1.0) and offline environment (1.9.3) are significantly different.

2. "Why 6.1.0 instead of 1.x.x?" — Because the `<release>` in `maven-metadata.xml` is set to 6.1.0 (this is the JUnit team's official "current latest stable"). From the JUnit team's perspective, 1.x.x is an update on the 1.x maintenance branch, not the "latest stable".

3. "Why the 6.x major instead of 5.x?" — JUnit merged the version numbers for console-standalone. The 5.x version number doesn't even exist (the 5 series is used by the JUnit API; platform 5 corresponds to platform 1.x; version numbers are unified after 6).

4. "Why trust <release>?" — This is the "latest release" pointer defined in the Maven standard, maintained by the JUnit team when publishing, and is much more reliable than the Solr interface on search.maven.org (the latter hasn't been updated in a year, which is the root cause of #1866).

chagong (approved):

PR #1887, opened 2026-07-14
feat: route real test program output to Test Results

wenytang-ms:
## Problem

Running a test with the Java Test Runner splits output across two disconnected surfaces:

- Program and debug output is delivered by java-debug as standard DAP `output` events, which normally appear only in the **Debug Console**.
- Test results are driven by the runner's socket protocol and displayed in **Test Results**.

The runner also echoed raw JUnit/TestNG control frames into Test Results, producing protocol noise such as `%TESTS`, `%TESTE`, and TestNG JSON frames instead of useful user-facing output.

See eclipse-jdtls/eclipse.jdt.ls#3834 for the broader discussion.

## Change

- **Mirror Debug Console output in Test Results.** Attach a `DebugAdapterTracker` to the test debug session and forward all non-telemetry DAP `output` events through `TestRun.appendOutput`.
- **Keep output at the test-run level.** DAP output is process-level and has no reliable test identifier, so it is not guessed onto individual test cases.
- **Correlate the exact debug session.** Add a unique internal run ID to the launch configuration instead of matching sessions by display name.
- **Remove control protocol noise.** JUnit no longer appends raw socket frames, and TestNG no longer echoes its JSON frames.
- **Preserve structured diagnostics.** Test states, stack traces, and expected-vs-actual values continue to use the Testing API. TestNG runner-level `error` frames are surfaced as structured test errors.

This requires no changes to java-debug; it consumes the standard DAP events already emitted by the debug session.

## Notes / limitations

- `TestRun.appendOutput` is an uncategorized stream, so stdout, stderr, and console color separation is not preserved.
- The Debug Console remains unchanged and continues to display the categorized output.

## Validation

- TypeScript compilation: clean
- ESLint on changed source files: clean
- Integration tests: 135 passing

### DIFF
--- a/src/runners/baseRunner/BaseRunner.ts
+++ b/src/runners/baseRunner/BaseRunner.ts
@@ -4,7 +4,7 @@
 import * as iconv from 'iconv-lite';
 import { AddressInfo, createServer, Server, Socket } from 'net';
 import * as os from 'os';
-import { CancellationToken, debug, DebugConfiguration, DebugSession, Disposable } from 'vscode';
+import { CancellationToken, debug, DebugAdapterTracker, DebugConfiguration, DebugSession, Disposable, ProviderResult } from 'vscode';
 import { sendError } from 'vscode-extension-telemetry-wrapper';
 import { Configurations } from '../../constants';
 import { IProgressReporter } from '../../debugger.api';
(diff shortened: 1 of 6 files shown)


wenytang-ms:
image
image

chagong (approved):

